PR #7, opened 2021-07-29
Adding msbuild+CMake+VS2019 CI build

DarkWanderer:
Adds a basic CI workflow which builds the project in both x86 and x64 platforms.
Example run: https://github.com/DarkWanderer/orbiter/runs/3196658204?check_suite_focus=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,51 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - 'Doc/**'
+      - 'Flights/**'
+      - 'Html/**'
+      - 'Images/**'
+      - 'Localdoc/**'
+      - 'Scenarios/**'
+      - 'Textures/**'
+      - 'Textures2/**'
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - 'Doc/**'
+      - 'Flights/**'
+      - 'Html/**'
+      - 'Images/**'
+      - 'Localdoc/**'
+      - 'Scenarios/**'
+      - 'Textures/**'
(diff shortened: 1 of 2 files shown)


hughesjs (approved):
It's a good start, would be good if we could get some testing integrated into the pipeline

SaadiSave:
> It's a good start, would be good if we could get some testing integrated into the pipeline

Are x64 builds supposed to fail?

DarkWanderer:
> Are x64 builds supposed to fail?

Yes, the codebase is not currently x64-compatible, but given it's the №1 item in TODO - and there are already a few people working on it - I left it in intentionally

SaadiSave:
So maybe run the x64 builds only for commits tagged with `x64`, so that the actions panel isn't full of failed builds?

DarkWanderer:
I'll let @mschweiger decide on that

SaadiSave:
> I'll let @mschweiger decide on that

Fair. It was just a suggestion.

mschweiger (commented):
Thanks for setting up the CI build. I have one question: will this try to build both the x86 and x64 versions? My master branch currently doesn't build x64. Could this be set up so that it only builds x86 when pushing to the master branch, but builds x64 when pushing to a different, x64-dedicated branch?

SaadiSave:
> Could this be set up so that it only builds x86 when pushing to the master branch, but builds x64 when pushing to a different, x64-dedicated branch? 

Yes, it could. The action just has to be set up to trigger upon push to the x64-branch.

mschweiger:
I've got a promising branch that might compile for both x86 and x64. About to merge into master, so I'll give the CI setup a try. Fingers crossed!

PR #10, opened 2021-07-30
OrbiterSDK now respects ORBITER_MAKE_DOC CMake option

matthew-reid:
Required to compile orbiter without docToPdf


mschweiger:
I think this has already been resolved in an earlier commit.

matthew-reid:
I see, it was fixed in 88c776a12694582a78885d9426bbc40f852f57dd. Closing.

PR #17, opened 2021-07-30
New Main Loop

TChapman500:
Don't worry, I kept the old loop behind a `#define DO_OLD_MAIN_LOOP` preprocessor directive.  Other than making the old loop more readable for me, I left it completely untouched.  The new loop processes all of the messages at once, then user input, then ticks the simulation as many times as needed to keep the simulation running at a realtime pace.  At least, that's what's supposed to happen.

### DIFF
--- a/Src/Orbiter/Orbiter.cpp
+++ b/Src/Orbiter/Orbiter.cpp
@@ -46,6 +46,7 @@
 #include "htmlctrl.h"
 #include "DlgCtrl.h"
 #include "GraphicsAPI.h"
+#include "Timer.h"
 
 #ifdef INLINEGRAPHICS
 #include "OGraphics.h"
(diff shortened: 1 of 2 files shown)


mschweiger:
Can you please explain what problem you are trying to solve with your edit? Note that Orbiter's fixed-timestep mode is not supposed to run in realtime. Its main point is to make the simulation independent of hardware by advancing the simulation by a defined time at each step, independent of how long it actually takes to compute. This is useful for debugging due to reproducibility independent of machine, or things like trajectory generation where you want to churn out data as fast as possible.

I am not convinced that the machine independence is preserved in your code, because the number of inner loop passes over the simulation steps is now dependent on system time, so the outer loop will be performed at different frequencies on different machines.

n7275:
Fixed timestep mode has been very valuable to me in some of my NASSP systems stability testing.

TChapman500:
I'm wanting a consistent simulation time step to run at realtime when a fixed time step is selected.  Also, I'm not sure I see what the problem with this code is.  I preserved the old loop behind a preprocessor directive, which allows fixed time steps to be used for debugging.

Here's what I'm trying to achieve:
https://www.gafferongames.com/post/fix_your_timestep/

n7275:
Wouldn't it be better to impliment as a 3rd timestep type option, rather
than a compiler option that modifies how fixed-step works?

TChapman500:
That should be relatively easy to do.

mschweiger:
It is an interesting concept, but it probably needs more thought. Note that Orbiter already does timestep subsampling in various situations, so there is a danger that you effectively end up with two nested subsampling loops, which would likely be bad news.

I may pull your PR into a test branch to play with it. Did you actually compile and try it? What were your observations?

TChapman500:
I'm getting about 2,000 compile errors right now, so I'm unable to test it out.

TChapman500:
Also, I'm under the impression that the tick subdivisions happens either in `BeginTimeStep()` or `EndTimeStep()` or `UpdateWorld()`.  I cannot find anything in the main loop itself that suggests subdivision.  So I do not believe that will be a problem.

PR #27, opened 2021-08-01
Adding 2-core parallel compilation

DarkWanderer:
* x86 compilation time down by 40-50% 10m -> 5-6m
* x64 compilation time down by 10-50% - 6m -> 3-5m
GitHub agents performance seems to be a bit inconsistent, but the improvement is definitely visible


mschweiger (approved):
I'll give it a try! Thanks for also spotting the missing dependency in ScriptMFD.

PR #39, opened 2021-08-03
Adding CMakePresets.json for default Visual Studio configs

DarkWanderer:
Allows to switch:

- x86/x64
- Debug/Release

Without restarting Visual Studio / regenerating solution. (need to open directory, rather than solution file, in VS)

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,4 +1,4 @@
-﻿{
+{
   "version": 2,
   "cmakeMinimumRequired": {
     "major": 3,
@@ -15,8 +15,8 @@
       "cacheVariables": {
         "CMAKE_C_COMPILER": "cl",
         "CMAKE_CXX_COMPILER": "cl",
-        "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}",
-        "ORBITER_MAKE_DOC": "OFF"
+        "ORBITER_MAKE_DOC": "OFF",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}"
       },
       "environment": {
         "VCPKG_FEATURE_FLAGS": "manifests,versions,binarycaching,registries"
@@ -24,7 +24,7 @@
     },
     {
       "name": "x86-debug",
-      "description": "Sets x86 arch, compilers, build type",
+      "description": "32-bit Debug build",
       "inherits": [ "base" ],
       "architecture": {
         "value": "x86",
@@ -41,7 +41,7 @@
     },
     {
       "name": "x64-debug",
-      "description": "Sets x64 arch, compilers, build type",
+      "description": "64-bit Debug build",
       "inherits": [ "base" ],
       "architecture": {
         "value": "x64",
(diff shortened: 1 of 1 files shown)


mschweiger:
Is this required for the CI run? I had deliberately left out the CMakeSettings.json file from the repository, since it contains user-specific settings, e.g. regarding building documentation etc. Also, I hate having the build directory inside the source tree, so my buildRoot and installRoot entries are different. I've also got a reference to the vcpkg toolchain file in there.

Is there an argument for including it in the repository?

DarkWanderer:
It is not required for CI (hence separate PR). Having this file committed to repository will allow developers to select from 4 configurations - x86-Debug, x86-Release, x64-Debug, x64-Release - in Visual Studio without having to:

- exit VS
- run cmake to regenerate projects/solution
- open Visual Studio again. 
 
This is purely a convenience feature for Visual Studio

DarkWanderer:
Added CMakePresets.json instead, see:

- https://docs.microsoft.com/en-US/cpp/build/cmake-presets-vs?view=msvc-160
- https://devblogs.microsoft.com/cppblog/cmake-presets-integration-in-visual-studio-and-visual-studio-code/

The idea behind this PR is to avoid having every developer setting up CMakeSettings.json for themselves (which might be tricky for some). `CMakePresets.json` has the advantage of storing default configuration + allowing custom `CMakeUserPresets.json` which advanced users can customize for themselves

DarkWanderer:
Closing to remediate local build issues

dbeachy1:
@DarkWanderer Is is possible to configure the defaults to put the build directory default to be outside the source tree (or maybe  just leave it blank?). If I understand @mschweiger correctly, the major concern with using defaults is that the build directory would be inside the source tree.

Regarding the other concern, what about even just leaving the default fields blank? The way, developers would still need to fill in settings in before building. I understand that sort of defeats the purpose of the presets file in the first place, but this would still allow developers to have all four build configurations (x86 Debug, x86 Release, x64 Debug, x64 Release) all in the same solution, if I understand correctly -- which, to be honest, is the big thing that I like :).

DarkWanderer:
What is the issue with having build artifacts inside the source directory, given that it's ring-fenced (all items are in subdirectories of `out/build`)? To be honest it feels strange to me to have build artifacts _outside_ source dir, given that basically any language build process produces something which is inside the source dir by default

DarkWanderer:
Let me rephrase this: I appreciate that this question is essentially matter of taste and do not claim that a particular way is unconditionally better. My suggestion comes from a standpoint that adding CMakePresets allows for new capabilities, while also helping to:

- streamline process for new contributors (they will not have to generate CMakeSettings.json)
- improve CI consistency (at the moment parameters are passed externally via command line)
- add capability for "remote" Linux development in future

In my opinion, having a default of building within source directory is the "least friction" option, while capability to build outside of source dir is retained, for cases where it is a concern

Of course, the decision is ultimately with you

dbeachy1:
FWIW, I don't feel strongly about this either way, and I agree it's a matter of personal taste. I'm going to close this PR for now since @mschweiger prefers the build settings as they currently are and this PR has been open for 22 days. We can always reopen this in the future if this resurfaces.

PR #55, opened 2021-08-04
Check install in CI regardless of branch/bitness

DarkWanderer:
Check 'install' in addition to 'build' in PRs (to make sure situations like [this](https://github.com/orbitersim/orbiter/runs/3242215571) do not repeat)


DarkWanderer:
PR checks fail because of #53

PR #61, opened 2021-08-05
speed up quicksaving

spacex15:
when you have a lot of scenarios saved
quicksave becomes very slow

this little change fixes the problem
moving the directory rescan just before dialog display


dbeachy1:
FWIW, this change looks fine to me -- the Orbiter launchpad is not visible anyway until the user exits the simulation, so it seems fine to defer updating the scenario list until the shutdown runs. @mschweiger what do you think?

mschweiger:
I am in the process of a major refactoring of the Launchpad class. Part of this is also cleaning up the scenario list update mechanism. This should resolve the problem addressed in this pull request. This has been implemented in branch https://github.com/orbitersim/orbiter/tree/99-launchpad_refactor, commit d452f09 . Can you please check if that branch solves the problems you were reporting? If so, I'll close this pull request.

spacex15:
It is ok for me, you deleted the refreshScenarioList from the save function

It will definitely solve the performance problem I had

Thanks

PR #64, opened 2021-08-06
ShuttleA work

GLS-SSV:
Made some corrections to the ShuttleA meshes, so everything fits together and looks better.
Also, fixed the vessel name being overwritten in the crew module, when more than 1 instance of the vessel was present.

There are more issues "inside" of the vessel (aux pod position switch loads "pressed" in VC, propellant displays overwrite/visual bug? in 2D panel, displayed propellant flow rate from tanks doesn't match displayed usage in thrusters, change to GDI logic to Sketchpad), but I just don't have the time to fix them right now... sorry.


mschweiger:
Thanks for this. Nice work on the meshes and the name display bug!

PR #65, opened 2021-08-06
Fix relevant pointer truncation warnings and one intermittent startup CTD for x64 builds

dbeachy1:
Was getting CTDs from a debug assertion failure on 50% of startups in x64 debug builds. See PR comments for details about the fix.

Also fixed all the other x64 pointer truncation warnings *except* for the ones in the DDE server code, which I believe is long-obsolete (if I'm remembering correctly).


dbeachy1:
Closed this PR because it turned out the x64 warnings and the intermittent startup CTD were already fixed in main. (I was testing in the trigger_sound_build_cmake branch).

Issue #35, opened 2021-08-02
Binary packages with CI


mschweiger:
I don't know if this is possible, and I certainly would need help setting it up, but here is what I want to do:

If CI builds Orbiter successfully after a pull to master, can it be set up to create a binary package and publish it on github? I have already set up cmake for a package target using cpack (both zip and msi, although zip is probably sufficient for this). A new package should replace the previous package, if any, otherwise I would probably fill up my storage allowance rather quickly. It only needs to create a package for the x64 build.

Any suggestions?

matthew-reid:
Sounds like that would work. You can create and delete releases via the [github REST API](https://docs.github.com/en/rest/reference/repos#releases).

If Github doesn't provide enough storage space, an alternative I currently use is [BackBlaze B2](https://www.backblaze.com/b2/cloud-storage.html) which gives you 10 GB free storage, and after that you only pay for what you use. You can use a REST API to manage files there too.

DarkWanderer:
I have a WIP branch which does just that:
https://github.com/DarkWanderer/orbiter/pull/4
Example build:
https://github.com/DarkWanderer/orbiter/actions/runs/1089362716
'Retention' can be set to 7-14 days and build refreshed once in that time.

Right now I'm fighting with an issue of binaries being in additional "Debug" directory. Once I clear that & switch to "release" build, I'll raise a PR to main repo

Xyon:
If retention is an issue we maintain a Digital Ocean "Space" within the Orbiter-Forum / OHM account which releases can be pushed to. It's inherently compatible with tools that work for AWS S3, too, so implementation is trivial; but I think github's asset storage is likely sufficient in the short term.

PR #72, opened 2021-08-08
Enable x86 artifact publishing

DarkWanderer:
To discuss. Suggested change per https://www.orbiter-forum.com/threads/x64-development.40026/post-585612
Potential risk: builds may "overflow" the quota


DarkWanderer:
Related: #35

DarkWanderer:
See #85

PR #75, opened 2021-08-09
Trigger_sound_build_cmake branch

dbeachy1:
Branch should be ready for merge now, if you agree it's ready.


mschweiger (commented):
All looks good to me. I did push another commit to the branch (before I saw the PR). This just added a user macro to the XRSound-lib-linktest project to make it easier to specify a different Orbiter build directory (which I had to do for my setup). I reverted the macro to the default location, so it should still work for you out of the box. If it is ok and you are happy I'll merge the branch.

PR #86, opened 2021-08-14
ElevationGrid Update.

jarmonik:
Changed ElevationGrid manager to return float instead of int16 for smoother and more accurate terrain collision matching. The physics engine does the interpolation in floating point format too.

I debugged a case long ago where surface elevation was defined in 0.1 meter resolution but the global lunar resolution is 0.5 meters, so, some precision is lost. It might be practical to let go the INT16 and move to float based elevation in general, in which case the planet specific scale factor would become obsolete. uint8 and int16 would remain valid tile formats of course.


jarmonik:
What is this "Failed Win32" check ? A failed fail check ?

mschweiger:
The 32-bit build reported an error:

```
D:\a\orbiter\orbiter\Src\Orbiter\surfmgr2.cpp(389,72): error C2664: 'void ElevationManager::ElevationGrid(int,int,int,int,int,int,INT16 *,float *,double *) const': cannot convert argument 8 from 'INT16 *' to 'float *' [D:\a\orbiter\orbiter\out\build\Src\Orbiter\Orbiter.vcxproj]
D:\a\orbiter\orbiter\Src\Orbiter\surfmgr2.cpp(389,68): message : Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast [D:\a\orbiter\orbiter\out\build\Src\Orbiter\Orbiter.vcxproj]
D:\a\orbiter\orbiter\Src\Orbiter\elevmgr.h(48,7): message : see declaration of 'ElevationManager::ElevationGrid' [D:\a\orbiter\orbiter\out\build\Src\Orbiter\Orbiter.vcxproj]
```

jarmonik:
It sounds like I have to start updating the DX7 engine. Is the DX7 OVP Client also included ?

jarmonik:
What is the current plan with the DX7 ? When are we going to leave it behind ? And what do we need to do before that can happen ?

jarmonik:
I'll just make two versions of the ElevationGrid function. That should do it.

PR #87, opened 2021-08-16
Break into debugger on fatal assertion failures if running under debugger

dbeachy1:
If Orbiter is running under a debugger, a fatal assertion failure will break into the debugger so the developer can snoop around and debug why the assertion failure occurred.


mschweiger (approved):
Sounds like a useful mod.

PR #95, opened 2021-08-18
Added integration tests

DarkWanderer:
* Added integration tests based on "Scenarios/Tests" directory
* Added `Catch` test framework - see https://github.com/catchorg/Catch2/blob/v3.0.0-preview3/docs/tutorial.md#top
* Had to touch command line parsing as current implementation does not correctly handle quotes: `"-s" "Tests\my_scenario"` causes it to hang


DarkWanderer:
Above successful build is bogus (tests were disabled) - closing until I can fix the segfault

PR #106, opened 2021-08-22
Moved XRSound-specific .gitignore section into separate file

DarkWanderer:


Sound/XRSound/.gitignore
```diff
@@ -0,0 +1,6 @@
+**/Debug/
```
dbeachy1, line 1: I'm curious what the reason for this change is. Am I correct to presume that the end result is the same? i.e., that Git will continue to ignore those files? But since should always be ignored by git in any build context, why does it need to change? Just want to be sure this won't break anything else.

dbeachy1: To elaborate further, there is already a `.gitignore` in the root of the repository that (correctly) contains this:

```
*.user
*.VC.db
**/.vs/
meshres*.h
CMakeSettings.json
out/**
!/Sound/XRSound/XRSound/**/*.vcxproj.user
**/*.bak
Sound/XRSound/**/Debug/
Sound/XRSound/**/Release/
Sound/XRSound/**/x64/
Sound/XRSound/**/dist/
Sound/XRSound/**/Debug-with-OrbiterRelease/
Sound/XRSound/**/Release-with-OrbiterRelease/
```
...and in my build configuration, at least, that is the file that is being used. So I'm curious what the purpose of the .gitignore settings in the CMake file is to begin with, because that is what could be affected by this change. Isn't the root `.gitignore` all we need anyway?

DarkWanderer: The records are _moved_ from the root .gitignore to subdirectory one - so that root file is not bloated by subproject-specific entries. This may be more convenient in the long run as more projects are added into tree (e.g. D3D9Client etc.). Very minor & suggestion only

dbeachy1: Ah, the `@@ -6,9 +6,3 @@ CMakeSettings.json` at the top of the change had me thinking this was in CMakeSettings.json. I see what you mean with wanting to spread  out the `.gitignore` settings to multiple files, but TBH I much prefer them all in one place just so everyone can find them. If we spread them out and people don't realize there are multiple `.gitignore` files in the tree, they can edit the wrong one and then it just gets confusing for everyone. Just my perspective, but I'd rather keep to a single `.gitignore` in the root.

Issue #103, opened 2021-08-21
Launchpad video tab: new behaviour


mschweiger:
Currently the video tab is only displayed in the server version if a graphics client is loaded.
Intended new behaviour:
- video tab always visible
- in the server version, display a combo box to select graphics client
- this is to replace the graphics client selection in the module tab
Reason:
- more intuitive and user friendly
- avoids ability to select multiple graphics clients, which currently is possible

mschweiger:
Closed with 103 video tab behaviour merge.

PR #110, opened 2021-08-23
103 video tab behaviour

mschweiger:


mschweiger:
Pick graphics client directly from video page.
Requires graphics client recompile.
closes #103

PR #112, opened 2021-08-24
#111: asan dll file name fixed for 32-bit builds

### LINKED ISSUE #111: AddressSanitizer: 32-bit build fails with ASAN enabled
This is due to a file name mismatch of the 32-bit asan dll: looking for clang_rt.asan_dynamic-x86_64.dll, but actual name is clang_rt.asan_dynamic-i386.dll.

mschweiger:
@DarkWanderer : re #107: Enabling AddressSanitizer failed for me for 32-bit builds, because the name of the asan dlls didn't match (VS 16.10.3). Can you confirm that you have the same problem with the original pull request (#107) and that this edit fixes it?

closes #111


DarkWanderer:
I have overlooked x86 buid completely - indeed, it breaks for me as well. I confirm this change fixes it

Thanks

PR #123, opened 2021-08-28
122 xrsound ci build

mschweiger:
Include XRSound build in CI build runs
- Download irrKlang package
- Cache package on CI server


dbeachy1:
Nice work on this @mschweiger ! 👍

mschweiger:
Thanks! Unfortunately it doesn't seem to work. XRSound is still not included in the build. I am investigating, but if anybody has an idea what the problem is, please let me know.

mschweiger:
Everything seems to work now. I checked that
- The CI build triggers a failure if XRSound fails to build.
- The cache is successfully hit, avoiding repeated downloads of the irrKlang libraries.

The cache remains valid for a limited time (7 days). If no cache hit occurs during this period, the cache is deleted, next time a new download is performed. So in the worst case, there would be a download once a week, but in practice much less frequently. That shouldn't put any undue load on the irrKlang download server.
@dbeachy1 : If you are happy, I'll merge the branch.

dbeachy1:
Great work, @mschweiger ! Looks good to me. 👍

dbeachy1 (approved):

PR #124, opened 2021-08-28
Fix for /fsanitize=address flag not being passed to correct targets

DarkWanderer:


dbeachy1:
@mschweiger This PR seems fine to me, but you know the Orbiter CMake flow better than I do.

PR #126, opened 2021-08-30
Test framework - WIP

mschweiger:
Merged from DarkWanderer:integration-tests

@DarkWanderer Thanks for working on the testing framework! I've merged your PR into a separate branch (test_framework) to give it a try before merging into main.

To start with, I've got a few questions:

1. I resolved a merge conflict in build.yml. Can you just check that it looks ok.

2. When running in console mode, Orbiter seems to hang on closing the console at the end of a second simulation session. To reproduce:
- Start orbiter, don't attach a graphics client.
- Start a scenario
- Type 'exit' in the console
- Start the scenario again
- Type 'exit' again
- Orbiter hangs.

Can you reproduce this? Can it be fixed?

3. In commit 828e310 I am hiding the console at the end of a session if Orbiter wasn't started from the console. Is this ok, or does it interfere in any way with the testing regime (e.g. do you need to keep the console open to see the test results?). It would be good if the console can be closed here to retain the previous behaviour.

I'll do some more testing and let you know if I come across anything else.


dbeachy1:
@DarkWanderer Did you catch @mschweiger's question above? Just making sure. 😄

DarkWanderer:
Hi. Yes, I've seen this question - haven't had time to look into it, apologies

Thanks for looking into this @mschweiger. On your questions:
> I resolved a merge conflict in build.yml. Can you just check that it looks ok.

Yes, the conflict resolution looks OK to me

> When running in console mode, Orbiter seems to hang on closing the console at the end of a second simulation session. To reproduce:

I have not reproduced the hang itself, but I have noticed that the first command typed after *re*starting simulation from launchpad is not read by ReadConsole. If you type `exit<enter>exit<enter>` - the command properly works. Is the behavior same in your case? Basically, the first command after restart is just skipped

Interestingly, it does not reproduce in case of launching from actual console - suggesting that there may be some interaction between `ShowWindow` and `ReadConsole`, or it's a side effect of `TerminateThread`. It is not caused by your change - the behavior is same in `main`. I'll look into it

> In commit 828e310 I am hiding the console at the end of a session if Orbiter wasn't started from the console. Is this ok, or does it interfere in any way with the testing regime (e.g. do you need to keep the console open to see the test results?). It would be good if the console can be closed here to retain the previous behaviour.

Yes, I do not see any issues with that

PR #128, opened 2021-09-03
Correcting Shipedit calculation of vertex normals

GLS-SSV:
Added missing normalization to vertex normals in Shipedit.


mschweiger:
I guess the scaling with icount can then be dropped since the normal vector is rescaled with the length anyway. Doesn't make a difference to the result, but saves a few processor cycles and looks cleaner.

PR #134, opened 2021-09-08
Fixes #133  - Fix problem with lua script loading.

### LINKED ISSUE #133: Lua scripts don't always load
The oapi_init.lua and scripts loaded with run() should explicitly specify ./ as the start directory.

tclarke:
Loading lua scripts from 'Script/' is not properly defined in stock lua and should explicitly specify the current directory as './Script'


mschweiger:
closes #133

PR #135, opened 2021-09-09
Fixes for building with community edition of VS (no MFC)

tclarke:
Include afxres.h
Add MFC detection and skip building Date and Shipedit utils when missing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,7 @@ find_package(OpenGL QUIET)
 find_package(HTMLHelp)
 find_package(Doxygen)
 find_package(LATEX)		
+find_package(MFC)
 
 # We don't query Qt with find_package because we need the 64-bit version here
 set(Qt5_x64_DIR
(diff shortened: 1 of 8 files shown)


dbeachy1:
BTW I'm using the VS 2019 Community Edition and Orbiter builds fine for me without these changes. Do we really need to silently skip building the Date and ShipEdit utilities if the developer building Orbiter does not enable the MFC option in their VS 2019 installer? The VS 2019 installer can be re-run at any time, not just on first-time installation.

tclarke:
I hadn't noticed the option for MFC. It's not completed obvious in the install and in older versions you needed the professional or enterprise to get the libraries. That said, it would be nice to disable those as they are the only thing that requires MFC and they aren't necessary if you're just doing development on the core. The change isn't silent, there are warnings written to the cmake log similar to the warning that disabled XRSound when you don't have the needed library.

JDat:
IF I understand correctly, than there is possibility to compile orbiter (almost) without MFC. If yes, than compiling without MFC helps to port Orbiter to linux. I like this idea while it not break existing code.

PR #144, opened 2021-09-22
Add DXSDK download into GitHub Actions

DarkWanderer:
This adds support of new D3D9Client build into GitHub Actions build

Example successful build: https://github.com/DarkWanderer/orbiter/actions/runs/1263239535


DarkWanderer:
^ @jarmonik

Issue #145, opened 2021-09-26
Deleting vessel cause crash when running with external graphic client


spacex15:
Testing on the d3d9client branch

The external graphic client clbkDeleteVessel is never called by orbiter core

I got a crash running in debug mode because of this

I suggest to change the following code in orbiter.cpp

#ifdef INLINEGRAPHICS
			oclient->clbkDeleteVessel ((OBJHANDLE)vessel);
#endif

to

#ifdef INLINEGRAPHICS
			oclient->clbkDeleteVessel ((OBJHANDLE)vessel);
#else
			if(gclient)
			    gclient->clbkDeleteVessel((OBJHANDLE)vessel);
#endif

same problem exists with other virtual module functions 

Or maybe the d3d9client  should register these functions ?

riojax:
Please, can you retest it?

PR #146, opened 2021-10-10
RFC: D3D9 client merge into `main`

DarkWanderer:
Suggest to review & discuss proposed merge of D3D9Client into Orbiter `main`. Full history of d3d9client has been merged, and CMake intergrated with main build. Potential points of contention: 
* API backward compatibility
* Increased size of build
* Documentation/includes placement

All credit of course goes to @jarmonik & other D3D9 contributors (I only helped to integrate it into Orbiter repo) - however, I think kicking off discussion on this early would be beneficial


spacex15:
I tested the d3d9client branch and the only problem I saw is issue #145

riojax:
I tested the d3d9client branch and the issues #157 and #145 are fixed.

dbeachy1:
Is it still the intention to merge D3D9 into Orbiter's main branch? Does this still need more testing after the merge conflicts are resolved?

derkweijers:
Lets get this moving shall we?
As proposed [here](https://www.orbiter-forum.com/threads/openorbiter-project-fork.40481/post-596198) I'd like to propose that we merge this into a fork that will include enhancements over the Orbiter core.

spacex15:
Martin, I see that you improved a lot the D3D9client CMakefiles.txt, 
Thanks a lot for that

I still do not understand this line
set(ORBITER_SOURCE_SDK_INCLUDE_DIR "${CMAKE_BINARY_DIR}/Orbitersdk/include" "${CMAKE_SOURCE_DIR}/Orbitersdk/include")

I do not see why d3d9client needs to search for "${CMAKE_BINARY_DIR}/Orbitersdk/include" ?
binary dir instead of source dir ?

I suggest to change to set(ORBITER_SOURCE_SDK_INCLUDE_DIR "${CMAKE_SOURCE_DIR}/Orbitersdk/include")

for me it causes build to fail when it is not a clean rebuild all

mschweiger:
@spacex15 you are right, it is a bit strange that both the source and binary directories are included. I don't think the line was written by me, so I am not sure what the idea here was. The build directory version of Orbitersdk/include may have been added to include dynamically generated header files.

Even stranger, there is also an "IncludeDir" which points to the _source_ Orbitersdk/include, but is used as an _output_ target for dynamically generated header files, which seems very unclean.

I'll try to figure out the intent here and see if I can come up with an alternative solution.

mschweiger:
@spacex15 : does this latest commit work better for you?

spacex15:
Yes it works for me
minor problem still persistent

It builds everytime the following D3d9client DLLs

DrawOrbits,TerrainToolbox,DX9ExtMFD,GenericCamera

even if nothing waas touched

PR #160, opened 2022-01-13
Fix 64bit compiler warnings

riojax:


jarmonik:
The DDE server might get removed from the orbiter if not then Martin has the knowledge of what it's supposed to do and ability to actually test it, so we should left that part of the code alone.  I can't verify if the changes are correct. Can you scrub the PR and make a new one for the first two commits.

riojax:
Ok, I fixed the DDE code to work again and also split the code to allow easily not compile it or remove it on a future, but I can understand this decision. I will make a new PR without those changes.

P.S. It's easy to check a DDE server using the freeware [DDEQuery](http://www.chrisoldwood.com/win32/ddecmd/ddecmd16r.zip) program, you don't need to have matlab, scilab or excel installed to check it.

JDat:
What about config option in Makefile to enable/disable include DDE server compilation when building Orbiter code?

riojax:
That was the idea on this PR, fix the actual DDE code and add a pre-processor flag to allow not compile this code, this flag will be set from cmake, but I think that this must to wait.

riojax:
If you want to review it, I did a [local branch with only this change.](https://github.com/riojax/orbiter/commit/572e5d59977bb0903179afde1a62beaccd9fd8ee)

PR #161, opened 2022-01-14
fix ShiftCG() in out-of-focus vessel

GLS-SSV:
This fixes the ShiftCG() call from an out-of-focus vessel shifting the vc click areas of the in-focus vessel, as reported in https://www.orbiter-forum.com/threads/orbiter-beta-r90-suspected-bug-with-shiftcg-call-and-vc-click-spots.40308/


jarmonik:
This modification is already included in d3d9client branch

PR #162, opened 2022-01-16
- Fix 64bit compiler warnings

riojax:
- Fix 64bit compiler warnings
- DrawAPI.h Add FMATRIX4 initialization
- Use proper encoding
- Click area only on focused vessels


jarmonik:
Is there a reason why some special charters are replaced with their ASCII code ?

riojax:
That's because modern MSVC compilers don't use Windows-1252 encoding by default, but this is the encoding that we need to render it properly, by that I hardcoded some characters on Windows-1252.
Other option is use the /source-charset and /execution-charset switches on CL.EXE, but this is unreliable and is prefered to hardcode it.

jarmonik:
Ok thanks. I wasn't aware that this kind of problem has emerged.

PR #164, opened 2022-01-17
Multiple bugfixes

riojax:
- Fix camera missing default case and a copy & paste error
- MFD sanity checks to avoid DBZ and BoF
- Minor fixes on Orbiter.cpp
- Fix race condition on Vessel.cpp
- Added initial values


jarmonik:
Is there a reason why sscanf return value is stored to "n" even if the return value is not used or needed ?

riojax:
Yes, that's due some compilers/runtimes do weird things if you don't get the return value on some functions, for example, modifying the default behavior. It's not very important if you know the compiler, but is a good recommendation to avoid problems (very difficult to debug) in a future.

PR #182, opened 2022-04-18
Search for plugin dependency DLLs in the plugin's subfolder

matthew-reid:
My plugin DLL has several dependency DLLs which I want to group under a subfolder to avoid polluting the Plugins folder. This pull request adds support to search for dependencies under module subfolders.

See also [discussion thread](https://www.orbiter-forum.com/threads/support-for-plugin-subfolder-dll-search-path.40449/) for this feature.

OrbiterModuleSubfolder
.


matthew-reid:
Closed due to issues raised in [forum thread](https://www.orbiter-forum.com/threads/support-for-plugin-subfolder-dll-search-path.40449/). A more secure and robust solution is required.

Issue #180, opened 2022-04-12
Button highlighting in Camera Dialog box is a bit weird


supersonic71:
Just a visual issue.

Steps - 
1 ) Go to Camera Dialog box and choose the "Track" tab
2) Select each mode one-by-one (Target, Absolute, Global, etc).
Only the first button is highlighted (blue border) even though I chose another button and I'm in a different camera mode now. 

Similar case
1 ) Go to Camera Dialog box and choose the "Track" tab
2) Select a button, then move cursor out of the dialog (maybe to adjust position with right mouse). Go back to camera dialog box. 
3) Repeat the same with all the buttons
For some reason multiple buttons are highlighted (blue border) even though only one camera track mode can be active at a time. 

Video - https://www.youtube.com/watch?v=5c4HOTCC-Kk

Tested on Open Orbiter.exe 32bit, DX7, Windows 11
Happens in Orbiter 2016 too

supersonic71:
Thanks for the fix @riojax !
Can confirm that first issue no longer occurs. 

Could you try the second part once? 

> 1 ) Go to Camera Dialog box and choose the "Track" tab
> 2) Select a button, then move cursor out of the dialog (maybe to adjust position with right mouse). Go back to camera dialog box.
> 3) Repeat the same with all the buttons
> For some reason multiple buttons are highlighted (blue border) even though only one camera track mode can be active at a time.

I think I'm still able to reproduce this one.

PR #186, opened 2022-04-30
Fix #179 #180 and limit warp at atmospheric flying

### LINKED ISSUE #179: occasional crash on scenario loading in D3d9client
I had an occasional crash in void Scene::SetupInternalCamera(D3DXMATRIX *mNew, VECTOR3 *gpos, double apr, double asp)
at first timestep

I suggest to add a test on camera.vnear on the following line in OVP\D3D9client\scene.cpp
// Something is very wrong... abort...
	if (Camera.hObj_proxy == NULL || Camera.vProxy == NULL ) return;

and change it to
// Something is very wrong... abort...
	if (Camera.hObj_proxy == NULL || Camera.vProxy == NULL || Camera.vNear == NULL) return;

riojax:
Apply fix #179
Remove button highlighting to address #180
Limit warp on atmospheric flying

### DIFF
--- a/OVP/D3D9Client/Scene.cpp
+++ b/OVP/D3D9Client/Scene.cpp
@@ -3332,7 +3332,7 @@ void Scene::SetupInternalCamera(D3DXMATRIX *mNew, VECTOR3 *gpos, double apr, dou
 	Camera.vNear = (vPlanet *)GetVisObject(Camera.hNear);
 
 	// Something is very wrong... abort...
-	if (Camera.hObj_proxy == NULL || Camera.vProxy == NULL) return;
+	if (Camera.hObj_proxy == NULL || Camera.vProxy == NULL || Camera.vNear == NULL) return;
 
 	// Camera altitude over the proxy
 	VECTOR3 pos;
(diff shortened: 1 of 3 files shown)


jarmonik:
Does the time warp code limit the warp factor on Moon ? If yes, then that's a failure. Using a dynamic pressure might be better idea than altitude. Also, a feature like this should be discussed in the Forums to gain more information about user preferences.

riojax:
Good idea, I will redo this part.

riojax:
> Also, a feature like this should be discussed in the Forums to gain more information about user preferences

Totally, the idea behind to limit this is avoid the simulations errors that I did see on atmospheric flying at more than x10 warp speed. This method only check when the user selects a new warp speed, by that it is not very "intrusive" and the performance impact is near none.

Anyway all feedback is pure gold, and I will post it on the forum.

maxq519:
_Please_ make this an optional feature!  I have a very fast machine and can handle 100x easily.

riojax:
Due feedback the warp limit on atmospheric flying will be an external module (shipped or not, I don't know for now)
Thank you all for your time :)

PR #193, opened 2022-07-16
Update Config.cpp - Change default MFD texture size for virtual cockpit

supersonic71:
I think the default res of 256 is too blurry in VC, so changed it to 1024. 
From http://orbiter.dansteph.com/forum/index.php?topic=14577.0
Comparison : http://www.dansteph.com/screenshots/200806-005943_FastStoneEditor1.jpg

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,7 @@ jobs:
 
     - name: Configure
       working-directory: ${{ github.workspace }}/out/build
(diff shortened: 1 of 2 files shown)


supersonic71:
Would also like to change these three defaults as suggested in the DanSteph guide - 
http://orbiter.dansteph.com/forum/index.php?page=welcome

1) Enable Cloud Shadows, Specular Ripples and Local Lights by default. 
2) Change MFD refresh rate from 1 to 0.1 
3) Set Focus Follow mouse to false. Both Windows and Ubuntu programs in general don't use focus follow. So I think that should be the default behavior for Orbiter too.

supersonic71:
Hi, friendly ping to maintainers

PR #199, opened 2022-08-13
Update Dragonfly.cpp oapiLoadMeshGlobal

supersonic71:
Default dragonfly scenarios are causing a "mesh missing" error, doing this edit seemed to fix it.
In Orbiter2016 Dragonfly mesh was in Meshes but in this Orbiter its in Meshes/Dragonfly.


mschweiger:
I have already implemented this fix in pull request #201 before reading this.
So I'm closing this - thanks anyway for the fix!

PR #235, opened 2022-09-05
Spacex15 patch xrsound usersound

spacex15:
fixes a crash 

The xrsound modification to include THGROUP_USER engine sound was missing a modification in vessel.h
in order to get the thruster group level


spacex15:
@dbeachy1 can you look at that ?
I am very sorry, I forgot to include this file in my pull request (vessel.h)

Causes a bad crash in the 64 bit version

dbeachy1:
@spacex15 Hold up here -- why do you have to modify an Orbiter core file just for XRSound? TBH I'm not comfortable with that, and it shouldn't be necessary.

dbeachy1:
@mschweiger could you please spot-check this commit?

Src/Orbiter/Vessel.h
```diff
@@ -546,8 +546,29 @@ class Vessel: public VesselBase {
...
+	inline double GetThrusterGroupLevel_User() const
+	{
+		double level = 0.0;
+		int    number = 0;;
+		ThrustGroupSpec* tgs;
+
+		for (DWORD i = 0; i < nthruster_grp_user; i++)
+		{
+			tgs = thruster_grp_user[i];
+			for (DWORD ii = 0; ii < tgs->nts; ii++)
```
dbeachy1, line 558: Martin can weigh in further, but I find `ii` easy to mix up with `i` when scanning the code in the inner loop here. What about naming the `ii` variable something like `thrusterIndex` so it's more descriptive?

spacex15:
@dbeachy1 
If you prefer, it is possible to implement the user engine sound without modifying orbiter core

special case in your EngineDefaultSoundPreStep::clbkPreStep to calculate thrust level using only existing vessel API

see proposed code below

should we do that ? 

thanks

// this is invoked at every timestep.  This code was moved here / ported from the EngineSoundsPreStep class in the XR codebase.
void EngineDefaultSoundPreStep::clbkPreStep(const double simt, const double simdt, const double mjd)
{
    double totalThrustLevel=0.0;

    // these sounds play even if the ship is not in focus
    VESSEL *pVessel = GetVessel();
    if (!pVessel)
        return;     // sanity check 

    if (m_thgType == THGROUP_USER)
    {
        THGROUP_HANDLE thgroup;
        DWORD groupcnt;
        DWORD totalnumber=0;
        
        DWORD indexthruster;
        DWORD thrustercnt;

        groupcnt = pVessel->GetUserThrusterGroupCount();
        for (DWORD indexgroup = 0; indexgroup < groupcnt; indexgroup++)
        {
            thgroup = pVessel->GetUserThrusterGroupHandleByIndex(indexgroup);
            thrustercnt = pVessel->GetGroupThrusterCount(thgroup);

            for (DWORD indexthruster = 0; indexthruster < thrustercnt; indexthruster++)
            {
                totalThrustLevel += pVessel->GetThrusterLevel(pVessel->GetGroupThruster(thgroup, indexthruster)); 
                totalnumber++;
            }
        }
        if(totalnumber > 0)
            totalThrustLevel /= totalnumber;
    }
    else totalThrustLevel = pVessel->GetThrusterGroupLevel(m_thgType);
    if (totalThrustLevel > 0)   // should sound be playing?
    {
        const float volume = XRSoundEngine::ComputeVariableVolume(m_minVolume, m_maxVolume, totalThrustLevel);

        // DEV DEBUGGING ONLY: sprintf(oapiDebugString(), "Engine thrust level = %lf, volume = %f", totalThrustLevel, volume);

        // OK if sound already playing here; if so, this call will merely change the volume
        PlayWav(true, volume);
    }
    else
        StopWav();
}

spacex15:
@dbeachy1 
no problem on my side, XRSound is your baby, I'll let you decide 

the idea was just to be similar to OrbiterSound in which there was a default engine sound

dbeachy1:
Thanks for the info for for your work on this, @spacex15. For what it's worth, unless @mschweiger feels your commit to the Orbiter core should be changed, I'd prefer to keep both your commits. But let's see what @mschweiger thinks first. 👍

mschweiger:
I'll have a look at this, but it will have to wait - I am just about to cross the Alps on a bicycle ;) I had a brief look at the vessel.h mods and there are a few things I don't quite understand.

dbeachy1:
No worries and no rush, have a great trip @mschweiger ! 😄

mschweiger:
Ok, back from holidays now. My concern with the code changes in vessel.h is this:

- the modification to GetThrusterGroupLevel means that the request for the level of _any_ user-defined group now simply returns an average of _all_ user-defined groups. What is the rationale for this? In what situation would the average over all user-defined thruster groups be useful, and how would the user now get the level of a specific user-defined group?

I gather this is somehow related to a requirement by a modification to the sound module, While I haven't looked at the sound mod, I would be inclined to doubt its validity if it requires this change to the core file. Why can't user-defined thruster groups be treated independently, like the default groups (e.g. main, retro, hover)? The thrusters within a group generally have a logical relationship, but this is not the case across groups, so bundling them together seems counterintuitive. A bit more explanation would be required.

dbeachy1:
Thanks for the info, @mschweiger -- I see what you mean, I didn't think of this before. It seems to me like adding default main engine sound as an average of all user-defined groups isn't particularly useful, so I propose just reverting the latest commit to XRSound if there are no objections.

mschweiger:
Thanks @dbeachy1 . No objections from me, but maybe wait a bit to give @spacex15 opportunity to comment.

spacex15:
thanks @mschweiger for your advice
I understand your objection 
may I suggest that you add a control preventing the api function to process getthrustergrouplevel with thgroup_user because it crashes

@dbeachy1  on the thgroup_user default sound the idea was to be iso functional with orbitersound and to facilitate the transition
I ll let you decide if you want to implement it or not
you can without the core modification
(see the code I suggested in the discussion thread)

again, thanks for the time you spent on this !

mschweiger:
@spacex15 : yes, you are correct. Thanks for pointing this out. I have opened a new issue #246 to deal with that problem.

PR #239, opened 2022-09-07
Implement Tesseral Gravity Perturbations

### LINKED ISSUE #231: Nonspherical Pertibations do not include tesseral terms
WIP implementation https://github.com/n7275/orbiter/tree/tesseralGravity
Discussion Thread: https://www.orbiter-forum.com/threads/improved-tesseral-gravity-model.40266/

n7275:
This PR implements a C++ version of Pine's singularity-free harmonic gravity.

Closes #231


mschweiger:
Thanks for your work on this! It would definitely be cool to have a complete perturbation model in longitude as well as latitude in Orbiter.
I checked out your pull request to try it out, but it doesn't look like the data files for running the tesseral perturbation code are included. How do I go about obtaining them? And will they be included in the final pull request, or will users have to obtain them themselves?

The one thing still missing is some documentation about this feature. The natural place for this to go would be the perturbations technote (Doc/Technotes/gravity.pdf). Would you be willing to join that paper as a co-auther and add another chapter explaining the new feature? I would suggest
- a paragraph outlining the method and how it improves on the previous simple zonal expansion (including references to the original papers)
- how to set it up in Orbiter (additional config file entries, data file format and location, etc.)
- and since you already have done quite some testing, maybe also include your test results and comparison with reference data.

n7275:
Thank you Martin,

I would be happy to add some documentation, that's actually the primary reason that this is still a draft pull request. I will add a writeup on how the implementation works, and how to use the feature, and I will include test data.

The data files for this feature are located here: https://pds-geosciences.wustl.edu/dataserv/gravity_models.htm (the ascii SHA models)

I am reasonably sure that we can include these, but I want to make sure they're properly credited.

It requires adding two lines to the planet/moon config file to which you want to add tesseral pertibations, e.g.:

```
GravModelPath = Config\\jgl165p1.sha
GravCoeffCutoff = 90
```
added to moon.cfg

I think GravCoeffCutoff should probably be fairly low by default (10?) by default. I am also planning on adding a directory specifically for gravity models, just to keep things organized.

n7275:
@mschweiger this should be ready for review now, and it should work out of the box for Mercury, Venus, Moon, Mars, Vesta, all you need to do to test is enable non-spherical gravity. The default cutoff is 10x10, so 110 coefficients; performance impact is virtually none, but scales as O(n²) with increasing order/degree and O(n) with each additional vessel.

PR #250, opened 2022-09-25
246 user defined thruster groups

### LINKED ISSUE #246: Vessel class: thruster-group related functions don't fully support user-defined groups
Some thruster group methods, e.g. Vessel::GetThrusterGroupLevel, crash if called for a user-defined thruster group (ID >= THGROUP_USER).

mschweiger:
- User-defined thruster groups (THGROUP_USER+x with x >= 0) should now be fully supported by Vessel class methods and thruster-group related API functions.
- Some cleanup of code in Vessel class; converted thruster and thruster-group related lists to std::vector and std::array
- GetThrusterGroupLevel called with THGROUP_USER should no longer crash (see #235)

Closes #246

### DIFF
--- a/Orbitersdk/doxygen/CMakeLists.txt
+++ b/Orbitersdk/doxygen/CMakeLists.txt
@@ -3,9 +3,14 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in ${CMAKE_CURRENT_BINARY_DI
 
 set(out ${ORBITER_BINARY_SDK_DIR}/doc/API_Reference.chm)
 
+file(GLOB_RECURSE doxysrc
+	${ORBITER_SOURCE_SDK_DIR}/include/*
+)
+
 set(api_ref_out "")
 add_custom_command(
 	OUTPUT ${out}
+	DEPENDS ${doxysrc}
 	COMMAND ${DOXYGEN_EXECUTABLE} Doxyfile 2> nul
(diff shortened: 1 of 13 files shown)


mschweiger:
@spacex15 : It would be helpful if you can check if this pull request solves the problem identified in #235 (GetThrusterGroupLevel crashes with THGROUP_USER).

spacex15:
@mschweiger I can confirm that calling GetThrusterGroupLevel(THGROUP_USER) no longer crashes and returns the thruster level for the first user group 
but I do not understand how to call GetThrusterGroupLevel for other id groups 
calling GetThrusterGroupLevel(THGROUP_USER+ groupindex )? (does not compile)
add another argument for the user group index ?

mschweiger:
@spacex15 : you are correct, since the THGROUP_USER member of the THGROUP_TYPE enum was the upper bound of the enum, no valid addition was possible. I have now added an additional member to the enum to provide an extended upper bound. This should now allow constructs like this:

```
THGROUP_TYPE MyThrusterGroup = (THGROUP_TYPE)(THGROUP_USER+1);
```

which then allows to use MyThrusterGroup anywhere a THGROUP_TYPE is expected. Does this work for you?

spacex15:
Yes @mschweiger , I tried GetThrusterGroupLevel((THGROUP_TYPE)(THGROUP_USER + indexgroup))
and it works now

thanks

Issue #234, opened 2022-09-05
Feature request : include release builds in artifacts


supersonic71:
This is to let more people who want to help with testing. 
Correct me if I'm wrong, but currently Github actions produce an Orbiter build which is "Debug" Version. 
So people who don't have Visual Studio installed can't run these artifacts as it requires some debug dlls (https://www.orbiter-forum.com/threads/skybolt-client-development.40057/page-5#post-599087). 
Would it be possible to also include a "Release" version in the artifacts? Should be faster to run too.

mschweiger:
I wasn't even aware that the artifacts were packing debug builds. The build script specifies configuration "RelWithDebInfo", but this is done on the build rather than the config step, which may be ignored by the Ninja generator.

For Ninja, the solution may be to add the flag
`-DCMAKE_BUILD_TYPE=Release`
on the config step.

Testing this will have to wait because I will be out of the loop for a few weeks. I'm just putting this comment here to remind myself.

PR #267, opened 2022-10-19
#234: Fix artefact output as debug build.

mschweiger:
build.yml: Added -DCMAKE_BUILD_TYPE=Release to Ninja configuration step and replaced RelWithDebInfo with Release for all config settings, to see if that fixes output artefacts being emitted as debug builds.


mschweiger:
Closes #234.

Issue #275, opened 2022-10-21
D3D9Client: fix incorrect rescaling of elevation mod tiles


mschweiger:
The tile scale and offset parameters of elevation mod tiles are incorrectly rescaled for target resolutions != 1. This leads to elevation artefacts where elevation modifications are active (see for example Mojave spaceport with installed Antelope Valley texture pack).

PR #276, opened 2022-10-21
#275: D3D9Client: Fix elevation artefacts for elev_mod tiles

mschweiger:
D3D9Client: apply tgt_res to the rescale and offset values of elevation mod tiles.
This should fix the elevation artefacts at Edwards and Mojave spaceport (Antelope Valley texture pack)


jarmonik (approved):
I don't remember what the idea was with the original code but let's proceed with your changes.

PR #293, opened 2022-11-03
Added the missing planet- & suface-base- axis display code to D3D9Cli…

computerex:
…ent.


schnepe2 (approved):
Looks like this is what **I wanted** but couldn't do 👍

PR #314, opened 2022-11-27
Update NRLMSISE-00 to latest release

riojax:
As @schnepe2 pointed on my latest PR, the best solution is to use the latest author release (https://git.linta.de/?p=~brodo/nrlmsise-00.git)

### DIFF
--- a/Src/Celbody/Vsop87/Earth/Atmosphere/EarthAtmNRLMSISE00/EarthAtmNRLMSISE00.h
+++ b/Src/Celbody/Vsop87/Earth/Atmosphere/EarthAtmNRLMSISE00/EarthAtmNRLMSISE00.h
@@ -12,6 +12,8 @@
 // MSIS atmosphere model implementation
 // ======================================================================
 
+extern "C" void gtd7(struct nrlmsise_input* input, struct nrlmsise_flags* flags, struct nrlmsise_output* output);
+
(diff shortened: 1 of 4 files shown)


schnepe2:
I would approve this PR :+1: (see [my review on the former -closed- PR](https://github.com/orbitersim/orbiter/pull/189#pullrequestreview-1039321787) ), but it's the doctors task to decide that ;)

mschweiger:
I'll check this out as soon as I find a bit of time. In part, this pull request simply seems to be undoing my edits of the library (anything marked MSyymmdd), but it's been so long that I can't remember why I did those edits in the first place. Some of them look like mere performance improvements, but since they would have been profiled on obsolete hardware they are probably no longer relevant.
Can you point me to the substantial changes in this PR (functional changes, bug fixes)? Also, did you do any performance comparisons between the current Orbiter implementation and your version (e.g. large number of vessels in Earth atmosphere at various altitudes), just to make sure that no significant performance changes have been introduced?

schnepe2:
The initial reason for this PR was a fix for buffer overrun on access of `meso_tn3[]` by @riojax.
I then proposed to "[...]just use the latest version of the lib as it fixes that bug among others[...]".
Reverting your (@mschweiger) changes was most likely not the intention 😁.

PR #324, opened 2022-12-11
D3D9 atmosphere remake (WIP)

jarmonik:
-Complete rework of atmospheric rendering and some updates to planet rendering in general.
-Fixed and re-enabled shader cache feature for faster loading.
-Added a Sun glare which can be disabled form setup and adjusted/tuned from graphics controls dialog.
-Added a enable/disable control for the irradiance/dynamic ambient light feature.
-Screen space depth/normal buffer is now available for future SSAO implementation.
-Fixed asteroid/comet rendering bug reported in the forums.


mschweiger:
I temporarily patched up the previous problem to get a bit further. I now get a message box `Failed to compile a shader - [...]\NewPlanet.hlsl(486,9): warning X3571: pow(f, e) will not work for negative f, use abs(f) or conditionally handle negative values if you expect them`.
Am I using a wrong version of the runtimes? I didn't have any problems with the previous D3D9 client version.

jarmonik:
Thanks for the reports. Looks like I forget to re-run the tests with lot of features disabled after making modifications. It's normal for the "pow" instruction to raise a warning if not secured from negative input. The runtimes should be correct if it gets that far, no changes in runtime requirements.

jarmonik:
Looks like there is a problem with shader cache. When a shader is loaded from a cache, feature on/off toggles have no effect since they are applied during compiling. How can I overlook something like this.

mschweiger:
The latest commits are fixing the reported issues for me. Thanks! For what it is worth, I have attached the Earth atmosphere settings I came up with, but I agree that there probably isn't a single configuration that works for everyone, since apart from subjective preferences this is also a matter of monitor/graphics driver characteristics and settings.
[Earth.atm.cfg.txt](https://github.com/orbitersim/orbiter/files/10208470/Earth.atm.cfg.txt)

OVP/D3D9Client/VPlanet.cpp
```diff
@@ -1662,37 +1393,35 @@ void vPlanet::SetMicroTexture(LPDIRECT3DTEXTURE9 pSrc, int slot)
...
+			sprintf_s(file_path, MAX_PATH, "Textures/%s", x.file);
+			
+			// If texture is not loaded, load it
+			if (MicroTextures.find(x.file) == MicroTextures.end()) {
+				if (D3DXCreateTextureFromFileA(pDev, file_path, &x.pTex) == S_OK) {
+					LogAlw("Microtexture [%s] loaded", x.file);
+					MicroTextures[x.file] = x.pTex;
+				}
+				else {
+					LogErr("Failed to read microtexture [%s] for [%s]", file_path, body.first);
```
schnepe2, line 1413: LogErr("Failed to read microtexture [%s] for [%s]", file_path, body.first.**c_str()**);
would be correct as body.first is std::string...

schnepe2:
Regarding the `ShaderCacheUse = 1|0` in D3D9Client.cfg.
When it's enabled, it creates the `Cache/Shaders` folder if not present. Is this the path we like to use? I mean The `Cache` folder[*].

There are several AddOns that would like to use such a folder for non-volatile storage. Currently only the root folder was mainly used for storing configurations or data-files.

I like the name `Cache`, but maybe there's a better name for this (soon to be filled with "garbage") folder?
`Temp` does not feel right either.

So this is just a Request For Comments.

---
[*] D3D9Client should maybe use a sub-folder like `Cache/D3D9Client/Shaders`, so the "user" of the files is more clear.
The guideline should then be something like `Cache\<AddonName>\`

OVP/D3D9Client/VPlanet.cpp
jarmonik: Thanks for the report. This should be fixed.

PR #334, opened 2023-03-31
PlanetarySystem cleanup/refactor

alektron:
- Replaces manually managed collections in `PlanetarySystem` (Vessels, Bodies, Celestials etc.) with std::vector and fixes singular/plural confusion with some of the member names.
I'm not a strong defender of 'Everything should use STL' but some of the re-/de-/allocation code was really verbose and hard to understand without any real advantage to a simple std::vector. It is not the first use of std::vector in the project so I assumed, it would be fine.
- Removes Orbiter.h as a dependency from `PlanetarySystem`.
`PlanetarySystem` is at the 'Base' of the simulation with higher level systems built on top of it. E.g. the render modules for...well...rendering or the Launchpad for Scenario selection, settings etc.
Having Orbiter.h (and therefore other dependencies like the Launchpad and a lot more) as a dependency of `PlanetarySystems` seems undesirable.
- Pulled TimeData out of Orbiter.h/.cpp into its own file to allow including its header without pulling in Orbiter. (see above).
- Attempt to get rid of some D3D dependencies where they were not strictly necessary. E.g. ShadowColor pulling in d3d.h just to return a D3DCOLOR where a Vector4 would suffice and the renderer can just convert it to D3D or whatever other backend.
- Attempt to refactor some D3D code that is needed for internal graphics and not easy to get rid of, so it is now at least neatly wrapped in INLINEGRAPHICS defines. This does not nearly cover everything but is just an attempt to one day maybe allow for a Server/NG build that does not need to compile against D3D headers.

This pull request consists purely of refactors, there SHOULD be no changes to functionality/features/behavior.
It was mainly a productive way for me to dive into the codebase for the first time.
Despite not offering any new features I believe some functions in `PSys` greatly improved in readability.
And getting rid of Orbiter.h in Psys.h should be a win in the project's structuring and maybe even improve compile times.

(This is my first ever pull request so please excuse any inexperience)

### DIFF
--- a/Src/Orbiter/Psys.cpp
+++ b/Src/Orbiter/Psys.cpp
@@ -7,10 +7,11 @@
 #include <stdio.h>
 #include <string.h>
 #include <io.h>
-#include "Orbiter.h"
+#include <algorithm>
+
 #include "Config.h"
 #include "Psys.h"
-#include "Astro.h"
+#include "TimeData.h"
 #include "Element.h"
 #include "Vessel.h"
 #include "SuperVessel.h"
@@ -18,24 +19,13 @@
 
 using namespace std;
 
-extern Orbiter *g_pOrbiter;
 extern TimeData td;
 extern bool g_bForceUpdate;
 extern char DBG_MSG[256];
 
-PlanetarySystem::PlanetarySystem (char *fname)
+PlanetarySystem::PlanetarySystem (char *fname, const Config* config, OutputLoadStatusCallback outputLoadStatus, void* callbackContext)
 {
-	name = 0;
-	nbody = 0;
-	nstar = 0;
-	nplanet = 0;
-	ngrav = 0;
-	nvessel = 0;
-	nsupervessel = 0;
-	//nuserlabel   = 0;
-	//labellist    = 0;
-	//nlabellist   = 0;
-	Read (fname);
+	Read (fname, config, outputLoadStatus, callbackContext);
 }
 
 PlanetarySystem::~PlanetarySystem ()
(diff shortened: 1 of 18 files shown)


Src/Orbiter/Psys.cpp
```diff
@@ -45,42 +35,34 @@ PlanetarySystem::~PlanetarySystem ()
...
-	if (ngrav) {
-		delete []grav;
-		grav = NULL;
-		ngrav = 0;
+	m_Name.clear();
+
+	//Vessel destructor broadcasts messages to every other vessel in 'vessels'.
+	//We remove it from the collection as soon as we deleted it to prevent the next Vessel to broadcast to the free'd one.
+	for (size_t i = vessels.size() - 1; i != static_cast<size_t>(-1); --i) { //Special reverse loop to prevent unsigned underflow
+		DelBody(vessels[i]);
```
schnepe2, line 44: Really `DelBody(...)` ? I would expect `DelVessel(...)` here

Src/Orbiter/Psys.cpp
```diff
@@ -45,42 +35,34 @@ PlanetarySystem::~PlanetarySystem ()
...
-	}
-	if (ngrav) {
-		delete []grav;
-		grav = NULL;
-		ngrav = 0;
+	m_Name.clear();
+
+	//Vessel destructor broadcasts messages to every other vessel in 'vessels'.
+	//We remove it from the collection as soon as we deleted it to prevent the next Vessel to broadcast to the free'd one.
+	for (size_t i = vessels.size() - 1; i != static_cast<size_t>(-1); --i) { //Special reverse loop to prevent unsigned underflow
```
schnepe2, line 43: Wouldn't this be even nicer?
```cpp
for (auto it = vessels.rbegin(); it != vessels.rend(); ++it) { // <= reverse iterator ...
	DelVessel(*it);
	vessels.pop_back();
}
```

alektron: It might be, however I can't guarantee if erasing items (pop_back) during reverse iteration will invalidate the iterator.
It SHOULD be fine, considering that the iterator is probably just an index internally but I assume that the standard makes no guarantees in that regard so I wouldn't take my chances.

schnepe2: I am pretty sure the reverse iterators work fine. At least when I test them ( https://godbolt.org/z/dEja7ceTv ) 
I used gcc though, but I bet msvc would behave similar 😀 

> [...]Iterators and references to the last element, as well as the [end()](https://en.cppreference.com/w/cpp/container/vector/end) iterator, are invalidated.[...]

from https://en.cppreference.com/w/cpp/container/vector/pop_back

On a side-note: I really love the cleanup/refactor of the code! Unfortunatly pull-requests are not pulled very frequent 😞

dimitry-ishenko: > Wouldn't this be even nicer?

@schnepe2 it wouldn't. According to [cppreference.com](https://en.cppreference.com/w/cpp/container/vector/pop_back) calling `pop_back()` invalidates iterator to the last element (ie, `rbegin()`) and the `end()` iterator.

This, however, might be nicer:
```c++
while (vessels.size()) {
    DelVessel(vessels.back());
    vessels.pop_back();
}
```

schnepe2: I am very sure you have tested this much more than I did =D 👍

schnepe2 (approved):
I very much like this refactoring! It's worth the changes.
Does this "approval" help?

Issue #338, opened 2023-05-17
C++ standard conformance


dimitry-ishenko:
Looking through the code I see things like:

* Identifiers starting with underscore (in violation of [[lex.name]/3](https://timsong-cpp.github.io/cppwp/n4868/lex.name#3))
* Use of `char*` instead of `std::string`
* Use of windows-specific functions instead of standard library
* Use of home-grown solutions instead of standard library

I also can't tell which version of C++ standard the code follows (other than the [D3D9 client](https://github.com/orbitersim/orbiter/blob/bf1e0fd6958c6eb9d3f6289794d96795029af407/OVP/D3D9Client/CMakeLists.txt#L7) and some [tileedit](https://github.com/orbitersim/orbiter/blob/bf1e0fd6958c6eb9d3f6289794d96795029af407/Utils/tileedit/qt/src/CMakeLists.txt#L4) utility).

As a first step I would recommend picking a standard for the project, so that contributors know not to exceed it. 

Luckily, MSVC (VS 2019 16.10) has a full support for the [C++20 core language](https://en.cppreference.com/w/cpp/compiler_support#cpp20) and also the [Standard Library](https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features). So, this should be pretty easy - pick C++20 as the standard for the project.

Next is, obviously, would be to incrementally update the code-base to conform to the standard, as well as replace any windows-specific functions with the STL ones.

This will also help make the code easier to port to other platforms, such as Linux and MacOS.

n7275:
If you are interested in working on this, it would be worthwhile to collaborate with https://github.com/TheGondos who was working on some of the same things, especially some of the Windows-specific stuff.

I would do this as incrementally as possible so that anyone that is working on longer-running projects has less to worry about by way of merge-conflicts.

dimitry-ishenko:
@n7275 yeah that sounds good. Unfortunately, I don't see any changes in that repo landing in here. Did I miss something?

dimitry-ishenko:
Another thing is the coding style. Or, rather, lack of it... 😅 Is there one?

maxq519:
Just remember, a lot of us aren't professional programmers.  Might keep that in mind when writing...

dimitry-ishenko:
@maxq519 neither am I 😄. I am talking about a simple document with things like:

* Use `.h` extension for header files and `.cpp` for source files.
* Use 4 spaces for indentation.
* Use `#define` guards in headers.
[...]

PR #341, opened 2023-05-26
Clean-up patches

dimitry-ishenko:
This is a first batch of clean-up patches discussed in #338 to enable C++20 standard conformance.

The purpose of these patches is to enable C++20 conformance of the existing code with minimal changes, so that further clean-up can be done as the next step.

The patches were purposely made very small and easy to review/accept, and will also aid any further bisects if necessary to find bugs.

### DIFF
--- a/Src/Orbiter/OrbiterAPI.cpp
+++ b/Src/Orbiter/OrbiterAPI.cpp
@@ -2560,7 +2560,7 @@ DLLEXPORT void InitLib (HINSTANCE hModule)
 
 		char *(*mdate)() = (char*(*)())GetProcAddress (hModule, "ModuleDate");
 		if (mdate) {
-			int Date2Int (char *date);
+			int Date2Int (const char *date);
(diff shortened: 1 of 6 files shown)


CaptainSwag101:
The change to the Date2Int() function parameters seems to have caused certain celestial bodies to throw errors while loading into a simulation. In no particular order, the offenders are Phobos, Deimos, Miranda, Ariel, Umbriel, Titania, and Oberon. They complain about a "procedure entry point not found" for the function Date2Int, which did not happen prior to this PR.

n7275:
This bug would've been discovered by opening the simulation once. Please make sure that we aren't introducing bugs.

The issue here is that those moon modules can't be rebuilt (because we don't have the source). They will need to be replaced.

The change is good, but we can't be breaking the Orbiter in the process. I am very happy to be a reviewer on these if you want.

dimitry-ishenko:
Sorry guys, that's my bad. I've reverted it on my end but forgot to push before submitting the PR. Will fix shortly.

dimitry-ishenko:
I've submitted PR #345 with the fix.

PR #342, opened 2023-05-31
Clean-up patches #2

dimitry-ishenko:
Here is the 2nd batch of clean-up patches discussed in https://github.com/orbitersim/orbiter/issues/338 to enable C++20 standard conformance.

The purpose of these patches is to enable C++20 conformance of the existing code with minimal changes, so that further clean-up can be done as the next step.

Most of these patches consist of adding `const` to variables point to string literals, and where that's not possible, adding explicit `char*` casts.


dimitry-ishenko:
Rebased to include the recent Date2Int() fix and ran a few scenarios to make sure things are working.

PR #343, opened 2023-06-04
Body: Some cleanup and refactoring

Cleverking2003:
This PR does some refactoring of the `Body` class, like using more modern C++ features and simplifying the logic a bit.

### DIFF
--- a/Src/Orbiter/Body.h
+++ b/Src/Orbiter/Body.h
@@ -26,12 +26,12 @@ public:
 	Body (char *fname);
 	// create a body from a config file
 
-	virtual ~Body ();
+	virtual ~Body () {}
 
 	virtual int Type() const { return OBJTP_GENERIC; }
 
-	inline char *Name() const { return name; }
-	inline const char* FileName() const { return filename; }
+	inline const char *Name() const { return name.c_str(); }
+	inline const char* FileName() const { return filename.c_str(); }
 
 	virtual const void *GetParam (DWORD paramtype) const { return 0; }
 
@@ -101,7 +101,7 @@ public:
 	virtual void RPlace (const Vector &rpos, const Vector &rvel);
 	// Set object position and velocity in parent coords
 
-	virtual void Update (bool force = false);
+	virtual void Update (bool force = false) {}
 	// Update object to current simulation time.
 	// Called between the body's module clbkPreStep and clbkPostStep API calls
 
@@ -126,15 +126,15 @@ public:
 	// This allows objects to initialise any D3DDevice objects
 	// that persist beyond the lifetime of the visual
 
-	virtual void DestroyDeviceObjects ();
+	virtual void DestroyDeviceObjects () {}
 	// Clean up device objects before the D3DDevice is destroyed
 	// Default action: Calls DestroyDeviceObjects() for all children
 
-	virtual void RegisterVisual (VISHANDLE vis);
+	virtual void RegisterVisual(VISHANDLE vis) { hVis = vis; }
 	// Called by a graphics client to notify the body of visual
 	// creation.
 
-	virtual void UnregisterVisual ();
+	virtual void UnregisterVisual() { hVis = nullptr; }
 	// Called by a graphics client to notify the body of visual
 	// destruction.
 
@@ -143,8 +143,8 @@ public:
 
 	inline const VISHANDLE *GetVishandlePtr() const { return &hVis; }
 
-	StateVectors *s0 = NULL;    // body state at time t0
-	StateVectors *s1 = NULL;    // new body state at time t0+dt during update phase
+	StateVectors* s0 { nullptr };    // body state at time t0
+	StateVectors* s1 { nullptr };    // new body state at time t0+dt during update phase
 
 	// Operations on updated state vectors (only accessible during update phase
 	// AFTER the object has been updated)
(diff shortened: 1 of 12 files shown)


Orbitersdk/include/VesselAPI.h
```diff
@@ -145,7 +145,7 @@ class OAPIFUNC VESSEL {
 	 * \return Pointer to vessel's name
 	 * \sa GetClassName
 	 */
-	char *GetName () const;
+	const char *GetName () const;
```
dimitry-ishenko, line 148: This is a step in the right direction, however:

1. Won't this potentially break some existing add-ons?
2. `const char*` is hardly "modern C++". The function should really return `const std::string&`.

Disclaimer: I'm just a random guy who likes Orbiter and knows a little bit about C++.

Cleverking2003: 1. Yes, that could break compatibility, I will probably remove this change (though I hope nobody _actually_ uses the return value as a mutable string).
2. Yes, it's better to use `string` but this would definitely break compatibility here.

dimitry-ishenko: > though I hope nobody actually uses the return value as a mutable string

You'd be surprised. Look at the countless `char*` casts in my PR #342.

Cleverking2003: Yes, there are quite a lot const correctness issues but most of them should be easy to fix because even though the code uses mutable pointers, it never actually changes the strings themselves (so you can just add the missing `const`).

dimitry-ishenko: You are right, but that wasn't my point.

Cleverking2003:
Reverted the change to the API function.

Xyon:
I looked away and forgot what was happening with this. Is it ready for merge?

Src/Orbiter/Body.h
```diff
@@ -163,39 +163,28 @@ class Body {
...
 
-	VISHANDLE hVis;      // visual identifier passed to messages (NULL=no visual)
+	VISHANDLE hVis { nullptr };      // visual identifier passed to messages (NULL=no visual)
 
-	const CelestialBody *cbody; // orbit reference body
+	const CelestialBody* cbody { nullptr }; // orbit reference body
 
-	void Setup ();       // initialise body with default params
+	std::string name { "" };          // object name
+	std::string filename { "" };
```
dimitry-ishenko, line 179: IMHO this should be an instance of [`fs::path`](https://en.cppreference.com/w/cpp/filesystem/path).

Src/Orbiter/FlightRecorder.cpp
```diff
@@ -96,7 +96,7 @@ void Vessel::FRecorder_Activate (bool active, const char *fname, bool append)
 		if (!append) FRecorder_Reset();
 		bFRrecord = true;
 		char cbuf[256];
-		sprintf (cbuf, "Flights\\%s\\%s.pos", fname, name);
+		sprintf (cbuf, "Flights\\%s\\%s.pos", fname, name.c_str());
```
dimitry-ishenko, line 99: If we ever want to make Orbiter cross-platform, which we do, all paths should be converted to use forward slash.

If you look under [Path normalization](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#path-normalization) section, it says:

> ## Canonicalize separators
> All forward slashes (`/`) are converted into the standard Windows separator, the back slash (`\`). If they are present, a series of slashes that follow the first two slashes are collapsed into a single slash.

dimitry-ishenko: Plus, this line should really be something like:
```c++
auto path = fs::path{"Flights"} / fname / name + ".pos";
```
and not a stupid `sprintf`.

NB: For this to work `name` would need to be am `std::string` and not `char*`.

Src/Orbiter/Planet.cpp
```diff
@@ -475,7 +475,7 @@ intptr_t Planet::FindFirst (int type, _finddata_t *fdata, char *path, char *fnam
 	switch (type) {
 	case FILETYPE_MARKER:
 		if (labelpath) strcpy (path, labelpath);
-		else           sprintf (path, "%s%s\\Marker\\", g_pOrbiter->Cfg()->CfgDirPrm.ConfigDir, name);
+		else           sprintf (path, "%s%s\\Marker\\", g_pOrbiter->Cfg()->CfgDirPrm.ConfigDir, name.c_str());
```
dimitry-ishenko, line 478: See above. Also applies to pretty much any other `sprintf`.

Src/Orbiter/Vesselstatus.cpp
```diff
@@ -175,7 +175,7 @@ bool Vessel::ParseScenarioLine2 (char *line, void *status)
 			Base *base = ((Planet*)vs->rbody)->GetBase (trim_string(cbuf));
 			if (!base) {
 				char cerr[1024];
-				sprintf (cerr, "Scenario parse error for vessel %s: base '%s' not found on body '%s'.", name, trim_string(cbuf), ((Planet*)vs->rbody)->Name());
+				sprintf (cerr, "Scenario parse error for vessel %s: base '%s' not found on body '%s'.", name.c_str(), trim_string(cbuf), ((Planet*)vs->rbody)->Name());
```
dimitry-ishenko, line 178: If you can wait for my PR to update c++ standard to c++20, then all non-path-related `sprintf`s can be replaced with [`std::format`](https://en.cppreference.com/w/cpp/utility/format).

dimitry-ishenko:
> I looked away and forgot what was happening with this. Is it ready for merge?

I am not the submitter, but IMHO there is quite a bit more room for improvement to make this code real modern C++.

Src/Orbiter/Body.h
dimitry-ishenko: Also, default constructor for `std::string` will create an [empty string](https://en.cppreference.com/w/cpp/string/basic_string/basic_string), so there is no need to add `{ "" }` afterwards.

dimitry-ishenko: And, it would be nice IMHO to suffix internal (private and protected) member variables with `_` to: (a) identify that they are internal, and (b) if you later want to expose it, you don't have to come up with a creative function name, such as `getName()` or something, and instead you can just add a function called `name()`.

Cleverking2003:
@dimitry-ishenko The problem with using `fs::path` right now is that Orbiter needs to use at least C++17 to have proper support for this feature, which doesn't seem to be the case. So we should probably postpone this change until we have C++20 fully enabled.

Cleverking2003:
@Xyon In my opinion this PR is ready for merge as it is because some of the suggested changes cannot be doen yet.

dimitry-ishenko:
@Cleverking2003 FYI you can now use c++20 features

PR #361, opened 2023-07-02
Replace non standard _itoa

TheGondos:
This PR replaces MSVC specific _itoa with standard sprintf.
Rational for not using std::to_string or equivalent : _itoa is usually used in C strings heavy contexts, mixing C++ strings in there would only clutter the code for no good reason. Feel free to disagree, I can change my mind if necessary.
Integer formating is size limited so I used snprintf only where a buffer overflow was possible.
Code in OVP/D3D* was not modified since it will obviously needs MSVC to compile anyway.

### DIFF
--- a/Src/Orbiter/DlgCapture.cpp
+++ b/Src/Orbiter/DlgCapture.cpp
@@ -54,8 +54,10 @@ BOOL DlgCapture::OnInitDialog (HWND hDlg, WPARAM wParam, LPARAM lParam)
 	EnableWindow (GetDlgItem (hDlg, IDC_COMBO2), prm.ImageFormat == 2 ? TRUE:FALSE);
 
 	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_RESETCONTENT, 0, 0);
-	for (i = 1; i <= 10; i++)
-	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_INSERTSTRING, -1, (LPARAM)itoa(i, cbuf, 10));
+	for (i = 1; i <= 10; i++) {
+		sprintf(cbuf, "%d", i);
+		SendDlgItemMessage(hDlg, IDC_COMBO2, CB_INSERTSTRING, -1, (LPARAM)cbuf);
+	}
 	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_SETCURSEL, prm.ImageQuality-1, 0);
 
 	return TRUE;
--- a/Src/Orbiter/DlgCapture.cpp
+++ b/Src/Orbiter/DlgCapture.cpp
@@ -12,6 +12,7 @@
 #include "Orbiter.h"
 #include "Resource.h"
 #include "Resource2.h"
+#include <string>
 
 extern Orbiter *g_pOrbiter;
 extern HELPCONTEXT DefHelpContext;
@@ -55,8 +56,7 @@ BOOL DlgCapture::OnInitDialog (HWND hDlg, WPARAM wParam, LPARAM lParam)
 
 	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_RESETCONTENT, 0, 0);
 	for (i = 1; i <= 10; i++) {
-		sprintf(cbuf, "%d", i);
-		SendDlgItemMessage(hDlg, IDC_COMBO2, CB_INSERTSTRING, -1, (LPARAM)cbuf);
+		SendDlgItemMessage(hDlg, IDC_COMBO2, CB_INSERTSTRING, -1, (LPARAM)std::to_string(i).data());
 	}
 	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_SETCURSEL, prm.ImageQuality-1, 0);
 
--- a/Src/Orbiter/FlightRecorder.cpp
+++ b/Src/Orbiter/FlightRecorder.cpp
@@ -20,6 +20,7 @@
 #include <io.h>
 #include <direct.h>
 #include <errno.h>
+#include <string>
 
 using namespace std;
 
@@ -315,9 +316,7 @@ void Vessel::FRecorder_SaveEvent (const char *event_type, const char *event)
 
 void Vessel::FRecorder_SaveEventInt (const char *event_type, int event)
 {
-	static char cbuf[24];
-	sprintf(cbuf, "%d", event);
-	FRecorder_SaveEvent (event_type, cbuf);
+	FRecorder_SaveEvent(event_type, std::to_string(event).data());
 }
 
 void Vessel::FRecorder_SaveEventFloat (const char *event_type, double event)
--- a/Src/Orbiter/FlightRecorder.cpp
+++ b/Src/Orbiter/FlightRecorder.cpp
@@ -316,7 +316,8 @@ void Vessel::FRecorder_SaveEvent (const char *event_type, const char *event)
 void Vessel::FRecorder_SaveEventInt (const char *event_type, int event)
 {
 	static char cbuf[24];
-	FRecorder_SaveEvent (event_type, _itoa (event, cbuf, 10));
+	sprintf(cbuf, "%d", event);
+	FRecorder_SaveEvent (event_type, cbuf);
 }
 
 void Vessel::FRecorder_SaveEventFloat (const char *event_type, double event)
--- a/Src/Orbiter/Vessel.cpp
+++ b/Src/Orbiter/Vessel.cpp
@@ -41,6 +41,7 @@
 #include <iomanip>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string>
 
 #ifdef INLINEGRAPHICS
 #include "VVessel.h"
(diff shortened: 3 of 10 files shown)


Src/Orbiter/DlgCapture.cpp
```diff
@@ -54,8 +54,10 @@ BOOL DlgCapture::OnInitDialog (HWND hDlg, WPARAM wParam, LPARAM lParam)
...
 	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_RESETCONTENT, 0, 0);
-	for (i = 1; i <= 10; i++)
-	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_INSERTSTRING, -1, (LPARAM)itoa(i, cbuf, 10));
+	for (i = 1; i <= 10; i++) {
+		sprintf(cbuf, "%d", i);
+		SendDlgItemMessage(hDlg, IDC_COMBO2, CB_INSERTSTRING, -1, (LPARAM)cbuf);
+	}
 	SendDlgItemMessage (hDlg, IDC_COMBO2, CB_SETCURSEL, prm.ImageQuality-1, 0);
 
 	return TRUE;
```
dimitry-ishenko, line 63:
```c++
SendDlgItemMessage(hDlg, IDC_COMBO2, CB_INSERTSTRING, -1, (LPARAM)std::to_string(i).data());
```

Src/Orbiter/FlightRecorder.cpp
```diff
@@ -316,7 +316,8 @@ void Vessel::FRecorder_SaveEvent (const char *event_type, const char *event)
 void Vessel::FRecorder_SaveEventInt (const char *event_type, int event)
 {
 	static char cbuf[24];
-	FRecorder_SaveEvent (event_type, _itoa (event, cbuf, 10));
+	sprintf(cbuf, "%d", event);
+	FRecorder_SaveEvent (event_type, cbuf);
```
dimitry-ishenko, line 320:
```c++
FRecorder_SaveEvent(event_type, std::to_string(event).data());
```
And drop `static char cbuf[24];` altogether.

dimitry-ishenko: Or better yet, update the `Vessel` functions, so we don't have to come back to it again:
```diff
-	void FRecorder_SaveEvent (const char *event_type, const char *event);
+	void FRecorder_SaveEvent(std::string_view event_type, std::string_view event);

-	void FRecorder_SaveEventInt (const char *event_type, int event);
+	void FRecorder_SaveEvent(std::string_view event_type, int event);

-	void FRecorder_SaveEventFloat (const char *event_type, double event);
+	void FRecorder_SaveEvent(std::string_view event_type, double event);
```

Src/Orbiter/Vessel.cpp
```diff
@@ -2711,7 +2711,7 @@ bool Vessel::Undock (UINT did, const Vessel *exclude, double vsep)
 		for (n = n0; n < n1; n++) {
 			if (len >= 255) break;
 			if (n > n0) cbuf[len++] = ' ';
-			_itoa (n, cbuf+len, 10);
+			sprintf(cbuf+len, "%d", n);
 			len += (n < 10 ? 1 : n < 100 ? 2 : 3);
 		}
 		FRecorder_SaveEvent ("UNDOCK", cbuf);
```
dimitry-ishenko, line 2717: This whole block can be replaced with:
```c++
auto buf = std::to_string(n0);
for (n = n0 + 1; n < n1; ++n) buf += " " + std::to_string(n);
FRecorder_SaveEvent ("UNDOCK", buf.data());
```

dimitry-ishenko: Alt:
```c++
std::string buf;
for (n = n0; n < n1; ++n) buf += " " + std::to_string(n);
FRecorder_SaveEvent ("UNDOCK", buf.substr(1).data());
```

dimitry-ishenko:
@TheGondos please don't use `sprintf`. Its days in c++ are numbered. With c++20 we got the [Formatting library](https://en.cppreference.com/w/cpp/utility/format) which is the replacement for it.

But, even without the Formatting library, in many cases you can use `std::to_string::data()` as long as you using it in the same expression and not storing it.

Or, rewrite small bits of code to switch completely to `std::string`. The code actually becomes simpler.

Src/Orbiter/Vessel.cpp
dimitry-ishenko: Actually, the original code has several bugs in it:
1. It assumes a ship has 1-, 2- or 3-digit dock IDs (Elysium, anyone?). 
2. Ships undocking more than 88 docks at the same time will need more than 256-char buffer.
3. It allows buffer overflow, because it checks `if (len >= 255) break;` only after the fact of overflow. 🤦🏻 

I know these are all very unlikely events, but such vulnerabilities could be easily targeted by hackers.

TheGondos:
> @TheGondos please don't use `sprintf`. Its days in c++ are numbered. With c++20 we got the [Formatting library](https://en.cppreference.com/w/cpp/utility/format) which is the replacement for it.
> 
> But, even without the Formatting library, in many cases you can use `std::to_string::data()` as long as you are using it in the same expression and not storing it.
> 
> Or, rewrite small bits of code to switch completely to `std::string`. The code actually becomes simpler.

I'll take a rain check on that one. There was a similar issue in NASSP and trying to use C++ constructs just snowballed out of control ; it was ultimately rejected in favor of keeping good old sprintf...
Let's face it, C++ is not the most appropriate tool for working with strings. It took 20 years to introduce the glorified pointer+length that is a string_view, the very object that does not guarantee your char pointer will reference a NUL terminated string, rendering it completely useless if you need to pass it to standard functions.
Is C better? Probably not, but at least the code is already there.
This PR is about portability, not C++ guideline shininess. If the core team wants to go in that direction, that's fine, I'll adapt it ; but for now I'll wait for more comments ;-)

dimitry-ishenko:
@TheGondos I understand. And, you are right, we definitely don't want it to snowball out of control. I just figured it would be better to use `std::to_string()` since at some point we will have to switch to `std::string` anyway.

But, it's really good to see you contributing here. I was really happy to see your Linux fork, and I am sure you know more than anyone else the hoops we'll have to jump through to make it cross-platform. 💯 

BTW, please also add D3D9 client in your PR. With the DXVK library we can make it run natively on Linux.

Src/Orbiter/FlightRecorder.cpp
TheGondos: OK for the first part, but I won't change the API for now

Src/Orbiter/Vessel.cpp
TheGondos: ok

TheGondos:
Thanks for the comments, I've taken them into account where they don't imply changing everything ^^

dimitry-ishenko:
@TheGondos this is OT, but since we are here: Are you developing under Linux or Windows? I've been using msvc under wine, but I was thinking of switching to clang-cl. Just curious how you are doing it.

TheGondos:
> @TheGondos this is OT, but since we are here: Are you developing under Linux or Windows? I've been using msvc under wine, but I was thinking of switching to clang-cl. Just curious how you are doing it.

I'm currently working on Windows, I don't want to mess something up when I commit^^

dimitry-ishenko:
@TheGondos FYI with 32-bit build I get this:
```
Y:\orbiter\Src\Orbiter\Mfd.cpp(1150): warning C4477: 'sprintf' : format string '%lld' requires an argument of type '__int64', but variadic argument 1 has type 'int'
Y:\orbiter\Src\Orbiter\Mfd.cpp(1150): note: consider using '%d' in the format string
Y:\orbiter\Src\Orbiter\Mfd.cpp(1150): note: consider using '%Id' in the format string
Y:\orbiter\Src\Orbiter\Mfd.cpp(1150): note: consider using '%I32d' in the format string
```
because `INT_PTR` in 32-bit mode is `int` and not `int64_t`:

https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types#INT_PTR

dimitry-ishenko:
Also, please give some love to the OVP clients:
<pre>
OVP/D3D7Client/D3D7Enum.cpp:                strcat(cbuf, <b>_itoa</b>(dupeIDCount, idStr, 10));
OVP/D3D7Client/VideoTab.cpp:    SetWindowText (GetDlgItem (hTab, IDC_VID_WIDTH), <b>_itoa</b> (data->winw, cbuf, 10));
OVP/D3D7Client/VideoTab.cpp:    SetWindowText (GetDlgItem (hTab, IDC_VID_HEIGHT), <b>_itoa</b> (data->winh, cbuf, 10));
OVP/D3D7Client/VideoTab.cpp:            (LPARAM)<b>_itoa</b> (usebpp, cbuf, 10));
OVP/D3D7Client/VideoTab.cpp:        (LPARAM)<b>_itoa</b> (bpp, cbuf, 10));
OVP/D3D9Client/VideoTab.cpp:        if (<b>_itoa_s</b>(data->winw, cbuf, 32, 10)) return;
OVP/D3D9Client/VideoTab.cpp:        if (<b>_itoa_s</b>(data->winh, cbuf, 32, 10)) return;
OVP/D3D9Client/VideoTab.cpp:            if (<b>_itoa_s</b> (h, cbuf, 32, 10)) return;
OVP/D3D9Client/VideoTab.cpp:            if (<b>_itoa_s</b> (w, cbuf, 32, 10)) return;
</pre>

dimitry-ishenko:
> Yes, I can work around the 32bit warning but I think the root cause is using an INT_PTR to represent either an ExternMFD pointer or an MFD Id. Maybe it would be better to use a std::variant instead (I replaced it with a struct containing a union on my linux branch)

For now I would just do an ugly `#ifdef BUILD64` to handle the `sprintf` warning. Hopefully soon we can get rid of all `sprintf`s.

TheGondos:
> > Yes, I can work around the 32bit warning but I think the root cause is using an INT_PTR to represent either an ExternMFD pointer or an MFD Id. Maybe it would be better to use a std::variant instead (I replaced it with a struct containing a union on my linux branch)
> 
> For now I would just do an ugly `#ifdef BUILD64` to handle the `sprintf` warning. Hopefully soon we can get rid of all of them now that we are at c++20.

The Id can either be an ExternMFD pointer, or an integer between 0 and MAXMFD (12) for internal MFDs. This function is only valid for internal MFDs

dimitry-ishenko:
What you did in your PR was great.

PR #362, opened 2023-07-02
Use the standard thread library for the lua interpreter

TheGondos:
This PR replaces Windows specific thread handling with standard C++ constructs in the lua interpreter/console/MFD.
I've tested it on the DG atmospheric autopilot and solar system scenarios but I'd prefer to have some more feedback.
Is there some kind of stress test for the lua interpreter?


TheGondos:
Working on the failed test case, I think there is a bug in the core that is worked around with a TerminateThread,..

dimitry-ishenko:
@TheGondos did you see my comments in your _itoa PR?

TheGondos:
> @TheGondos did you see my comments in your _itoa PR?

Yes, I can work around the 32bit warning but I think the root cause is using an INT_PTR to represent either an ExternMFD pointer or an MFD Id. Maybe it would be better to use a std::variant instead (I replaced it with a struct containing a union on my linux branch)

dimitry-ishenko:
Sorry, since that PR was accepted I wasn't sure if you saw my comments. Let's continue in there.

TheGondos:
Since we are now using C++20, I took the liberty to switch to semaphores since they are more appropriate.
There was a deadlock when calling the Interpreter destructor in case its thread was already finished when trying to step it.
There was also a crash when calling exit from the intepreter thread so it now forwards its exit code to the main thread and exit is called from there.
CTests are now OK, I tried the DG atmo autopilot scenario several times in a raw and got no crash/freeze when exiting the scenario.

dimitry-ishenko:
All those late nights I've spent updating the code for c++20 are starting to pay off already 😅

TheGondos:
BTW, I tend to rebase/squash before merging because I find it easier to revert/cherry-pick if needed later. Tell me if it's not something you want.

Xyon:
> BTW, I tend to rebase/squash before merging because I find it easier to revert/cherry-pick if needed later. Tell me if it's not something you want.

Nah it's fine, I generally prefer to do that anyway unless the history on the branch is significant somehow

TheGondos:
Added a minor prototype change.
BTW, is there a way to run CTests in a loop? It's not really enough to declare some multithreading stuff OK over just a couple of tests. If I could launch it a few thousand times, it would be better...

DarkWanderer:
> Added a minor prototype change. BTW, is there a way to run CTests in a loop? It's not really enough to declare some multithreading stuff OK over just a couple of tests. If I could launch it a few thousand times, it would be better...

https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-repeat

`ctest --repeat-until-fail 1000 <other parameters>`

PR #365, opened 2023-07-10
Wrap std::min/max calls in parens

dimitry-ishenko:
This keeps Windows min/max macros from kicking in and causing errors.

### DIFF
--- a/Orbitersdk/include/DrawAPI.h
+++ b/Orbitersdk/include/DrawAPI.h
@@ -226,12 +226,12 @@ namespace oapi {
 #endif
 		float MaxRGB() const
 		{
-			return std::max( r, std::max(g, b));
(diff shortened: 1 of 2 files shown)


TheGondos:
> This keeps Windows min/max macros from kicking in and causing errors.

Aren't the min/max macros now disabled by default?

dimitry-ishenko:
> Aren't the min/max macros now disabled by default?

Not for addon dev who roll their own projects and simply include `DrawAPI.h` and/or `VesselAPI.h`.

CaptainSwag101:
NASSP addon dev here, the removal of min/max results in thousands of build errors when trying to build for OO any time we call any DrawAPI or VesselAPI function. If there's a better way to leave your commit intact while also fixing the build, please let me know!

PR #366, opened 2023-07-10
Replace itoa with std::to_string

TheGondos:
Second round to replace _itoa calls in OVP.
I took this opportunity to work around a warning introduced earlier in the 32bit build.


Src/Orbiter/Mfd.cpp
```diff
@@ -1147,7 +1147,7 @@ bool Instrument::FindScnHeader (ifstream &ifs) const
 	switch (id) {
 	case 0: strcpy (header+10, "Left"); break;
 	case 1: strcpy (header+10, "Right"); break;
-	default: sprintf (header+10, "%lld", id + 1); break;
+	default: sprintf (header+10, "%lld", (int64_t)id + 1); break;
```
dimitry-ishenko, line 1150: Even better than my ugly `#ifdef` idea 😄

PR #374, opened 2023-07-17
Use std::filesystem instead of windows API in Orbiter core

TheGondos:
This PR mainly replaces the findfirst/findnext windows API with std::filesystem iterators.
We could go further with introducing fs::path in more places but it will need to wait since it'll likely break the ABI.
Impacts where something may have been broken :
- scenario editor
- atmosphere modules enumeration/configuration
- video module enumeration
- module enumeration
- scenario enumeration
- Quicksave folder cleanup from the launchpad
- markers/features enumeration
- bases enumeration
- flight recorder (directory cleanup when overwriting a record)
One thing I've not been able to test confidently is the PlanetarySystem::ScanLabelLists function.

### DIFF
--- a/Sound/XRSound/src/Utils/FileList.cpp
+++ b/Sound/XRSound/src/Utils/FileList.cpp
@@ -53,56 +53,25 @@ void FileList::Scan(const char *pPath, const int recursionLevel)
 
     // This code was broken out from XRPayloadClassData::InitializeXRPayloadClassData().
 
-    WIN32_FIND_DATA findFileData;
-    char pConfigFilespecWildcard[MAX_PATH];
-    sprintf_s(pConfigFilespecWildcard, "%s\\*", pPath);  // iterate through all files and directories
-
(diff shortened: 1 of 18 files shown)


n7275:
@TheGondos I think this commit getting merged has caused a few build errors, but hopefully those should be easy to correct.

This happens locally on my end too, but: https://github.com/orbitersim/orbiter/actions/runs/6936782339/job/18869594903


https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c3533?view=msvc-170

I tried to find a quick fix, but I'm not seeing anything obvious. I think it must just be a compiler of cmake setting though.

Xyon:
Yeah I saw this build fail and meant to take a look at it, but I haven't had chance to. I don't think it's a major thing, but the build did succeed in the pre-merge checks.

n7275:
Yeah, I figured it passed the pre-merge checks, and obviously it worked in
for Gondos on their end. So it's gotta be a very simple fix, like a missing
local change. I'll look a bit more tonight.

Issue #378, opened 2023-07-25
Raytraced VC textures with panel lights


jarmonik:
Issue or a project to implement ray-traced textures for virtual cockpit

https://www.orbiter-forum.com/threads/raytraced-vc-textures-with-panel-lights.40347/

PR #379, opened 2023-07-25
Ray-traced VC Textures

jarmonik:
Implementation of ray-traced VC textures. Issue #378
https://www.orbiter-forum.com/threads/raytraced-vc-textures-with-panel-lights.40347/

**Work in progress. Do not Merge**


n7275:
Hi @jarmonik, regarding your "do not merge" comment, what we often do for NASSP dev, is leave PRs as "draft", until we're ready to merge them. Then we mark them as ready for review. That way they can't accidentally get merged early, but we can still use them as a placeholder for an ongoing project.

jarmonik:
This project is ready enough for shader development to start. Waiting feedback..

n7275:
@jarmonik, what is needed for a test of this feature?

jarmonik:
There is currently nothing to test. The shader would need to be finished first. Felix24 knows how to build ray-traced textures and how to write shaders so he knows both sides of the equation. I don't know who else could do that. I suppose I could try to write some kind of shader but I have close to zero experience with 3D modeling and ray-tracing, so, I am not going to make any textures.  It's a holiday season so it might take some time until things move forward. Meanwhile I try to improve the sun shadow quality.

n7275:
hi @jarmonik will this feature support local light sources in the VC as well as the baked lightmaps and sunlight?

jarmonik:
Felix24 might be better qualified to answer the question but based on my understanding it would be just baked lights and sunlight. Support for dynamic local lights is already there but it's unknown if the shader will use them. But I suppose that dynamic lights could still be used to lit dynamic objects inside the cockpit like a floating object of some sort. Even if the dynamic lights wouldn't effect on a baked mesh. 

We probably should add some kind of ambient light probe in the cockpit to detect the amount of sunlight shining through windows which would then dynamically increase the ambient light level inside the cockpit. It's not great if the cockpit is totally dark even if sunlight is coming in from a window. Some shaders should also include the local lights in addition to sunlight.

n7275:
> Felix24 might be better qualified to answer the question but based on my understanding it would be just baked lights and sunlight. Support for dynamic local lights is already there but it's unknown if the shader will use them. But I suppose that dynamic lights could still be used to lit dynamic objects inside the cockpit like a floating object of some sort. Even if the dynamic lights wouldn't effect on a baked mesh.
> 
> We probably should add some kind of ambient light probe in the cockpit to detect the amount of sunlight shining through windows which would then dynamically increase the ambient light level inside the cockpit. It's not great if the cockpit is totally dark even if sunlight is coming in from a window. Some shaders should also include the local lights in addition to sunlight.

I am so bad at remembering to reply to things. I will reach out on the forum.

I like the idea of the ambient light probe. It's amazing just how bright the sun is, from a lighting standpoint: https://www.flickr.com/photos/projectapolloarchive/21315590814/in/album-72157659042210300/

Who else has written shaders in the past?

Jordan-64:
> > Felix24 might be better qualified to answer the question but based on my understanding it would be just baked lights and sunlight. Support for dynamic local lights is already there but it's unknown if the shader will use them. But I suppose that dynamic lights could still be used to lit dynamic objects inside the cockpit like a floating object of some sort. Even if the dynamic lights wouldn't effect on a baked mesh.
> > We probably should add some kind of ambient light probe in the cockpit to detect the amount of sunlight shining through windows which would then dynamically increase the ambient light level inside the cockpit. It's not great if the cockpit is totally dark even if sunlight is coming in from a window. Some shaders should also include the local lights in addition to sunlight.
> 
> I am so bad at remembering to reply to things. I will reach out on the forum.
> 
> I like the idea of the ambient light probe. It's amazing just how bright the sun is, from a lighting standpoint: https://www.flickr.com/photos/projectapolloarchive/21315590814/in/album-72157659042210300/
> 
> Who else has written shaders in the past?

I think the photo from Flickr has a longer exposure time, you can see that because the camera was moved while taking the photo.

colinmcglothlin:
> We probably should add some kind of ambient light probe in the cockpit to detect the amount of sunlight shining through windows which would then dynamically increase the ambient light level inside the cockpit. It's not great if the cockpit is totally dark even if sunlight is coming in from a window. Some shaders should also include the local lights in addition to sunlight.

This is a good idea, and is basically what I'm trying to do. I currently grab the blue channel of gSun.Color as a proxy for the amount of sunlight present in the VC and use it to change the brightness of the ambient light illumination. gSun.Color is white in day and black at night, with some sunrise/sunset color mixing in between. I grab the blue channel because it is the last channel to increase in brightness at sunrise, and the first to decay at sunset, and allows for a more gradual transition between dynamic range regimes.

This does not take into account flying upside-down with the sun not visible through the windows. In this case, currently there is no change in the ambient light.

I want be able eventually to have multiple ambient light maps for sunlight coming from different directions. One with the light coming in from the forward direction, one for each side, and one for above.

colinmcglothlin:
> hi @jarmonik will this feature support local light sources in the VC as well as the baked lightmaps and sunlight?

This is a great idea. I'm not sure how to implement and test this though.

jarmonik:
The C++ side should be now in a working condition. The shaders still need some work. Feedback required regarding fully baked configuration.  

Ambient light probe experiment was successful, although, without HDR the small difference between dim local light and sun light causes some issues. With HDR upgrade, performance optimizations and quality improvements, this can work. With this upgrade it would be possible to block Earth glow coming in a closed payload bay.

It might be good idea to wait for the release of Orbiter 2024 before merge and then start updating the stock vessels since no more DX7 compatibility is required.

jalexb88:
@jarmonik I have been testing with internal dynamic shadows using code from this commit: a193053e2a2196646d38fb1e0152fa458aac8b48 The result in NASSP is quite good in my opinion: https://www.dropbox.com/s/emczp6g854riys4/Screenshot%202025-01-07%2020.36.06.png?raw=1 The ambient lighting may be a little too low now but before it was much too high in my opinion as the sun seemed to illuminate the entire interior face directly opposite to it, even if there is no window in between to shine through, whereas now it is just through an actual window. Maybe eventually some other way can be found to restore some of the lost ambient light in a more realistic way if that makes sense. Anyway I have built my own version of the Orbiter 2024 D3D9 client that has it enabled but would it be feasible to have this feature pushed to OpenOrbiter, maybe with an tick box to enable/disable in the options?

MrResiduals:
Is it possible to enable the sunlit VC shadows by default, instead of through the debug menu?

jarmonik:
> Is it possible to enable the sunlit VC shadows by default, instead of through the debug menu?
In a future it will be automatically enabled if the VC is configured to support shadows. Right now it's still in early state of development.

maxq519:
Would it be possible to have the state of the checkbox for enabling the shadows be saved, so that once it is enabled manually that preference is saved for that vessel and mesh?

n7275:
> Looks like GitHub fails to fetch sub-module/project [https://github.com/Microsoft/DirectXMath ](https://github.com/Microsoft/DirectXMath) How is that supposed to be done properly ? It's all **inline** module. No module specific binaries/libraries are being build build.

Possibly a question for @Xyon

jarmonik:
> "cmake magic", probably. It doesn't look like that's the most recent build issue, is the package pull still a problem?

Right now the problem is that Vulkan SDK doesn't found. I would expect that major SDK's being installed by default on server side and shared for every project (in which case it would be search path issue) but if that's not the case then Vulkan SDK would need to be installed on GitHub and made available for the project. I have no idea how is that supposed to work.

There's no rush with this. I can do development and it build's fine locally. But dependency problems should be solved before merge to main.  There's a lot of work to do before that,

PR #384, opened 2023-08-26
Add General Purpose "Airfoil" Definition

n7275:
This pull request is add a general-purpose "airfoil" as new, 4th type to Orbiter's existing AIRFOILSPEC type. This function CreateAirfoil4, like CreateAirfoil3 allows addon developers to specify their own functions for computing force and moment coefficients, however it expands greatly in the following areas:

Problem Statement:

CreateAirfoil3 allows the creation of multiple, static aerodynamic force-producing bodies in the local coordinates of the vessel to which they are bound. These Airfoils must be either "Vertical" like a wing, or "Horizontal" like a vertical stabilizer. In response to Mach and Reynolds number, velocity, angle of attack, and slip angle, Orbiter internally computes (by means a function passed from an addon by pointer) the C<sub>L</sub> , C<sub>D</sub> and C<sub>m</sub> coefficients. which Orbiter internally uses to apply aerodynamic forces and moments to a vessel. This presents several major limitations:

- Force coefficients C<sub>L</sub> and C<sub>D</sub> produce forces relative to airspeed vector, not vessel body axes.
- Aerodynamic Forces and Moments can only be functional dependant on AoA and Slip, but cannot be a function of both at the same time.
- At very high slip angles AoA becomes poorly defined, and because lift is dependant on AoA, lift force becomes poorly defined. In a spacecraft this force is often the _largest_ at high slip angles.
- Roll moments due to slip (dihedral effect, and other cross-coupling are not possible.

Proposed Solution is Provided in the following PR:

- Allow Orbiter to calculate axial, normal, and side force coefficients: C<sub>A</sub> , C<sub>N</sub> and, C<sub>Y</sub> and roll, pitch and yaw moments C<sub>l</sub> , C<sub>m</sub> and, C<sub>n</sub> from a user-defined function and allow _these_ six coefficients to be used in calculating forces and moments on the vessel.
- Follow the paradigm of previous Orbiter airfoil functions, which will allow free combination of this "Airfoil" (Spacefoil?) type with all other types and control surfaces.

Significant Benefits:
- Allow for the implementation of analytical complex models of aerodynamic coefficient functions which may exist in historical documentation, without hacks.
- Facilitate easy implementation of lookup table aerodynamics.

Implementing something like this for all six coefficients would not currently be possible:
image

This update has not been thoroughly tested yet. I would also like some input from addon devs.


jarmonik:
I haven't had time to study the implementation but it looks good to me, let me know when it's ready for merge.

n7275:
> I haven't had time to study the implementation but it looks good to me, let me know when it's ready for merge.

The physical simulation part of this is complete. I wanted to add a side-force visual helper vector to make debugging/development easier, since we have Lift, and Drag visualizations, but no way to see the force that's mutually orthogonal to them. I already have a good start on this, and most of what I have left is adding a check-box and checking the state of it in code.

I saw you mention "Orbiter 2024" on the forum, and wanted to make sure this made it in, so I'll be trying to get this finished up as quickly as possible. :)

n7275:
Side force now displays properly!
image

jarmonik:
This looks good to me, I can merge it or should we ask Martin to review this ?
Also the information given in the pull request should end-up into the documentation. In a cases like this I would probably just drop a pdf in the Technotes folder.

n7275:
I think this can be merged. I've tested a simple implimentation, and functionally it workes exactly like CreateAirfoil3 but more generalized so that hacks don't need to be done in addon code. I just wanted someone other than me to look over the code before it gets merged. I know @indy91 has as well.

The Doxgen comments should build with the API reference PDF.

I am more than happy to write up a guide on how to impliment an aerodynamic model using this method.

jarmonik:
Everything looks good to me. If a problem arise it can always be fixed later on.

jarmonik:
> I am more than happy to write up a guide on how to impliment an aerodynamic model using this method.
It would be nice to have some kind of a guide.

PR #385, opened 2023-09-16
Add gravity model for Earth; fix documentation.

n7275:
Add EGM96 model created by @Ajaja

https://www.orbiter-forum.com/threads/improved-tesseral-gravity-model.40266/post-609309


n7275:
@Xyon or anyone else, this should be okay to merge.

PR #388, opened 2023-10-27
Add system requirements to readme.md

### LINKED ISSUE #350: Add system requirements
I noticed Orbiter has no system requirements except for ram and disk space, so if somebody wants to know if their PC is enough powerful to run it, the only way is to install it.

There is a requirements table:

|  | Minimum requirements | Recommended requirements |
| ---- | ---- | ---- |
| RAM: | 4GB | 8GB |
| CPU: | Dual Core |  |
| GPU: | 200 GFlops | 500 GFlops |
| Disk: | 4GB of free space | 10GB of free space |

Is it ok?

Miner34dev:
closes #350

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ Orbiter is now published as an Open Source project under the MIT License (see
 D3D9Client graphics engine is licensed under LGPL, see [LGPL](./OVP/D3D9Client/LGPL.txt)
 
 ## Installation
+|  | Minimum requirements | Recommended requirements |
+| ---- | ---- | ---- |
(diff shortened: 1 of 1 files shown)


n7275:
It's fairly common for people to install the high-res planet textures. Maybe the recommended hard drive space should reflect this. Iirc they're on the order of 80GB total.

maxq519:
I'd make the recommended specs reflect the hi-res textures, but leave the minimum at like 5 GB

Miner34dev:
Fixed.

Issue #389, opened 2023-11-22
Build Failing because of auto


n7275:
@TheGondos I think this commit getting merged has caused a few build errors, but hopefully those should be easy to correct.

This happens locally on my end too, but: https://github.com/orbitersim/orbiter/actions/runs/6936782339/job/18869594903

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c3533?view=msvc-170

I tried to find a quick fix, but I'm not seeing anything obvious. I think it must just be a compiler of cmake setting though.

_Originally posted by @n7275 in https://github.com/orbitersim/orbiter/issues/374#issuecomment-1822100886_

jarmonik:
As far as I can tell it's NOT been downgraded. It was raised to C++20 without authorization and some commits been made that require C++20.  I would hope to focus on development instead of babysitting.

jarmonik:
The build didn't fail a week ago, so what changed.

n7275:
I think the root cause of this is: some pull requests waited a long time
for merge and became de-synced from upstream changes.

The build checks should have alerted us to this issue, but they didn't.

I think we need to pull, build and test each PR before merge.

The C++20 change was 5 months ago, was on a single PR and was merged
through normal approval channels. Gondos std::fs change depended on it,
this is what broke things. We either need to move to 20 or revert this and
leave it on hold for now.

I spent some time reading old forum posts, github emails and discord
messages. As far as I can tell and what I remember, we decided to hold off
on major API breaking changes at least until we make one "release". But
just updating the standard shouldn't break anything?

More thoughts later when I've had more coffee...

jarmonik:
> I spent some time reading old forum posts, github emails and discord messages. As far as I can tell and what I remember, we decided to hold off on major API breaking changes at least until we make one "release". But just updating the standard shouldn't break anything?

Yes, we are supposed to release one more Orbiter with DirectX 7 and then we can remove the DX7 parts and cleanup some dirty hacks we have needed to keep the D3D9Client working. After this is done I am still expecting the API to be binary compatible with existing add-ons. So, nothing major is expected to happen. There is no point to turn the API upside-down just in a heck of it. I am no mind-reader but I guess that most developers would appreciate **stability** regarding the API

dimitry-ishenko:
> As far as I can tell it's NOT been downgraded.

It was with this commit: c4c7be503a8bbc0ef1e3bd24c39534ca66ca9fae

[...]

PR #390, opened 2023-11-25
*Temporarily* Revert "Use std::filesystem instead of windows API in Orbiter core"

n7275:
This reverts commit acb91358197736b024a1136ada09a7868f8d9e68. *Temporarily*

I have built and tested that Orbiter functions normally.


n7275:
Closes #389

PR #393, opened 2023-12-30
Adding exports check

DarkWanderer:
One of current issues with Orbiter development is ensuring binary compatibility with Orbiter 2016 release

This PR adds a few workflow commands which generate a list of functions exported from Orbiter.exe and then diff them with exports of Orbiter 2016 edition. This allows to view changes in ABI against established baseline


jarmonik:
I don't understand much about these workflows and artifacts but it looks fine to me. Go for merge ?

PR #398, opened 2024-01-03
exclude non-64-bit-modules from 64 bit builds

### LINKED ISSUE #243: Physics Instability/Ephemeris issue?
This issue is most noticeable when taking off from Mars in some of the default Delta Glider or Shuttle A scenarios. Reference: https://github.com/Piraxus/OrbiterSkyboltClient/issues/8

I have noted similar issues randomly taking off from KSC, where the earth appears to slowly fall away from the vessel. Occasionally, semi-major axis will change as though there is a force being applied when no thrust is active.

Video: https://youtu.be/B3S6ZECeomk

I am not sure what the cause could be here, and I don't want draw the wrong conclusion.

Observations:
-Later MJDs seem to be better (or it might be a coincidence in what I observed).
-The "jumping" effect seems to be present only on Mars
-Sometimes it appears as though a phantom force is being applied to the vessel or planet (apparent acceleration).
-Reloading a scenario in Mars orbit, sometimes results in a drastically different Orbit.

Anything I can help test to make it more clear what the root cause is, let me know.

### LINKED ISSUE #353: Mars atmospheric model seems broken in 64bit build
Steps to reproduce
1. Use a 64bit build
2. Disable all perturbations (on Physics settings)
3. Enable Complex flight model (on Vessel settings)
4. Select scenario: Delta-glider -> Brighton Beach
5. Select ship: GL-02
6. Apply & Close
7. Go to the cockpit and click on HROZ LVL button

P.S. This issue was happening before PR #243

n7275:
This is a temporary fix to make 64-bit builds usable "out of the box" without commenting out anything in sol.cfg

closes #243
closes #353 

This is not a replacement for the missing modules. It just prevents CMAKE from packaging non-working 32-bit DLLs with OOx64

### DIFF
--- a/Src/Celbody/Ariel/CMakeLists.txt
+++ b/Src/Celbody/Ariel/CMakeLists.txt
@@ -28,6 +28,7 @@ set_target_properties(${CELBODY}
 	FOLDER Celbody
 )
 
+if(NOT CMAKE_CL_64) #remove me when this celbody has source code that can be built
 # Installation
 install(PROGRAMS
 	${CELBODY}.dll
@@ -36,3 +37,4 @@ install(PROGRAMS
 install(DIRECTORY Config
 	DESTINATION ${ORBITER_INSTALL_ROOT_DIR}
 )
+endif()
--- a/Src/Celbody/Deimos/CMakeLists.txt
+++ b/Src/Celbody/Deimos/CMakeLists.txt
@@ -27,7 +27,7 @@ set_target_properties(${CELBODY}
 	PROPERTIES
 	FOLDER Celbody
 )
-
+if(NOT CMAKE_CL_64) #remove me when this celbody has source code that can be built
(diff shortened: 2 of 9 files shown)


jarmonik:
Let me know when this is ready for merge.

Src/Celbody/Ariel/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set_target_properties(${CELBODY}
 	FOLDER Celbody
 )
 
+if(NOT CMAKE_CL_64) #remove me when this celbody has source code that can be built
```
DarkWanderer, line 31: We have `BUILD64` macro in root `CMakeLists.txt` - worth using that for consistency

n7275:
Will be fixed by #630

Issue #375, opened 2023-07-18
Establish a C++ version for the project


riojax:
It's a good idea to set a specific C++ version and follow it to avoid future problems, for example, someone pushing C++20 code that breaks all MSVC 2019 builds.

Personally I propose to follow C++11 that is very compatible between a lot compilers/versions.

dimitry-ishenko:
Why would you establish a version that's over 10 years old? All respectable compilers (GCC/Clang/MSVC) support c++20, and that's the version that's been established for the project:

https://github.com/orbitersim/orbiter/blob/09baae4050b07c5a3edd31da4e2680e8cb2eb952/CMakeLists.txt#L9-L11

riojax:
[You pushed](https://github.com/orbitersim/orbiter/commit/632f1c4f8eebe5508891ac196458f7b1574efa03) those lines in CMakeLists.txt and also pushed some C++20 code, but there are not a public discuss about to switch to C++20 for this I opened this issue to do it properly, in an open and public manner giving voice to anyone.

About to use C++20 I am totally against it because it is not fully supported on MSVC 2019 and the implemented part have serious unfixed STL issues.

dimitry-ishenko:
> You pushed those lines in CMakeLists.txt and also pushed some C++20 code

Yes I did 😝 

> but there are not a public discuss about to switch to C++20, for this I opened this issue to do it properly, in an open and public manner giving voice to anyone.

I brought it up in an issue here and in the forums.

> About to use C++20 I am totally against it because it is not fully supported on MSVC 2019 and the implemented part have serious unfixed STL issues.

Upgrade your compiler. Believe it or not, but MSVC is currently **the one** compiler that has a full support for C++20:

https://en.cppreference.com/w/cpp/compiler_support#cpp20

Others are not far behind though.

> P.S. I wrote (and write) tons of C++03 portable to every arch/OS without major problems.

Why not C++98?... 😉 It's 2023... maybe it's time to upgrade?

n7275:
> I don't have forum access, and here... I think that I didn't see it, sorry. [Is this the issue?](https://github.com/orbitersim/orbiter/issues/338)

The forum and discord are still quite active. It is probably worth joining for the discussions that happen there.

dimitry-ishenko:
> Ok, this can be a problem then to port it to Linux or BSD (also the MSVC 2019 support is incomplete and bugged)
[...]

PR #402, opened 2024-01-05
Fix deprecated <strstream> (depricated since C++98)

schnepe2:
In addition to issue #375, replaced deprecated `<strstream>` by `<sstream>`
Depricated since C++98 (See: https://en.cppreference.com/w/cpp/header/strstream)


Src/Orbiter/Orbiter.cpp
```diff
@@ -16,7 +16,7 @@
 #include <stdio.h>
 #include <time.h>
 #include <fstream>
-#include <strstream>
+//#include <sstream>
```
robertrypula, line 19: Should it be commented out?

schnepe2: No, actually it should have been removed 😉 Since nothing in `Orbiter.cpp` is using any of that.
Unfortunately it was left as "commented out". I'll fix (remove) that!

PR #403, opened 2024-01-06
Fix Option Save

GLS-SSV:
Options in Visual settings and Physics settings pages were not being immediately saved, which caused any changes to be lost when changing to another options page. This PR should fix that issue, and also adds logic to disable the Planet night light level text box when Planet night lights are disabled, and also disable the terrain interpolation combo box when Surface elevation is disabled.

BEWARE: I'm using VS2017 and cannot compile the latest version of the code, so this was only tested in a 1 year+ old code version. I think it should work fine and not need any further work but still, it should be well tested in the latest code before merging.


n7275:
Probably wouldn't hurt to, pull, rebase, and compile just to make sure. But
yeah it's probably fine.

n7275 (approved):

n7275:
I rebased against upstream/Main and build on VS2022. No issues with the build, and the feature works as advertised. Approved!

Issue #410, opened 2024-01-11
oapiWriteScenario_float & oapiWriteScenario_vec have unpredictable precision


schnepe2:
As the ofstream object is passed as reference to several serialization methods and each one uses a different set of precision() modifiers (between 3 and 10!) the serialized output is not well behaved.

Current AddOns usually use a work-around like this (.\Src\Vessel\Atlantis\Atlantis\Atlantis.cpp):
```cpp
// --------------------------------------------------------------
// Write status to scenario file
// --------------------------------------------------------------
void Atlantis::clbkSaveState (FILEHANDLE scn)
{
	char cbuf[256];
	// ...
	sprintf (cbuf, "%0.4f %0.4f %0.4f %0.4f %0.4f %0.4f", arm_sy, arm_sp, arm_ep, arm_wp, arm_wy, arm_wr);
	oapiWriteScenario_string (scn, (char*)"ARM_STATUS", cbuf);
	// ...
}
```

Investigation showed, that during MFD serialization a **precision of only 2** was used.

I propose the following fix (a PR will come soon 😜):

1. precision is always 6 (it's the C++ default),
2. trailing zeros will be cut ("1.200000" => "1.2")
3. `oapiWriteScenario_float(...)` will always serialize with a decimal point ("1.0" will not become "1", but "1.0").
  This is not absolutely neccessary, but I think it's nice to see in the scenario file that this value was/is a floating point value...
5. the same rules also apply to vectors serialized by `oapiWriteScenario_vec(...)` - as they are just 3 `doubles`.

For info how it was discovered, [see this thread, too](https://www.orbiter-forum.com/threads/the-future-of-lua-with-orbiter-is-it-at-a-dead-end.41527/post-615019).

This should be 100% compatible with all older addons.
The only thing I can see is, that the serialization might take a tiny bit longer. The NASSP people might take a look, as their Scenarios are quit big when Orbiter is closed and ".\Scenarios\(Current state).scn" is created.

PR #411, opened 2024-01-11
Fix for #410 (unpredictable precision) 

schnepe2:
Fixes issue #410
Floating point output stream formatter added to serialize double/float values nice & tidy.


jarmonik:
Does this reduce the precision of things like vessel state vectors or (lng, lat) information in scenario files ?

schnepe2:
No, it doesn't.
Vessel state vectors might get a little better precision if anything would change.
It depended a lot on which MFD was set and/or what camara mode - hence unpredictable.
But never was it left any better than 6.

schnepe2:
I've looked at some "prominent" Vessel Addons to see whether they use oapiWriteScenario_float or `oapiWriteScenario_vec` but it seems almost all used the "`oapiWriteScenario_string` work-around" - as that was the only way to guarantee the output format accuracy/precision until now.
So: confidence is high! 😄

jarmonik:
Ok, Thanks.  Ready for merge ?

schnepe2:
Yes, ready.

PR #424, opened 2024-01-22
Use std::filesystem instead of windows API in Orbiter core

TheGondos:
This is a follow-up to PR #374 which failed to build after the C++ version was demoted.
It is the same as the original PR with the "auto &&" (abbreviated function template?) replaced with "std::function" parameters.

### DIFF
--- a/Sound/XRSound/src/Utils/FileList.cpp
+++ b/Sound/XRSound/src/Utils/FileList.cpp
@@ -53,56 +53,25 @@ void FileList::Scan(const char *pPath, const int recursionLevel)
 
     // This code was broken out from XRPayloadClassData::InitializeXRPayloadClassData().
 
-    WIN32_FIND_DATA findFileData;
-    char pConfigFilespecWildcard[MAX_PATH];
-    sprintf_s(pConfigFilespecWildcard, "%s\\*", pPath);  // iterate through all files and directories
-
-    HANDLE hFind = ::FindFirstFile(pConfigFilespecWildcard, &findFileData);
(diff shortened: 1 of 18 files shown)


spacex15:
Does not compile on visual studio 2022 64 bits

ifstream ulf(entry); in psys.cpp

Gravité	Code	Description	Projet	Fichier	Ligne	État de la suppression	Détails
Erreur (active)	E0289	aucune instance du constructeur "std::basic_ifstream<_Elem, _Traits>::basic_ifstream [avec _Elem=char, _Traits=std::char_traits<char>]" ne correspond à la liste d'arguments	Orbiter.exe (Modules\Server\Orbiter.exe) - C:\orbisource2\orbiter\Src\Orbiter\Psys.cpp	233

jarmonik:
That's odd. I did test compile it on VS2019 before merge (x86). Have you tried x86 ?

TheGondos:
Couldn't find a single compiler that accepts this on compiler explorer, so it must work only on VS2019 :sadface:
Managed to get clang/gcc to compile it by replacing the offending line with this :
`ifstream ulf(entry.path());`
@spacex15 : since I cannot test it (using VS2019 here), can you confirm it works on your side so I can push the fix?
Thanks

dimitry-ishenko:
> Couldn't find a single compiler that accepts this on compiler explorer, so it must work only on VS2019 :sadface:

@TheGondos https://godbolt.org/z/jqdKen91d

Your fix of adding `.path()` after `entry` is correct.

spacex15:
@thegondos I confirm the fix works on visual studio 2022

thanks

PR #425, opened 2024-01-24
Add support for VS2022

TheGondos:
This PR fixes the build issues when compiling with VS2022, it also adds CI jobs to detect potential future failures.
The artifacts for the VS2022 builds are not saved for now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,9 @@ jobs:
       fail-fast: false
       matrix:
         architecture: [x64, x86]
+        os: [windows-2019, windows-2022]
       
-    runs-on: windows-2019
+    runs-on: ${{ matrix.os }}
 
     steps:
     - name: Checkout
@@ -96,6 +97,7 @@ jobs:
       run: git diff -U0 --ignore-cr-at-eol --ignore-space-at-eol --no-index exports.2016.txt out/install/Orbiter/exports.txt
 
     - name: Upload exports
+      if: ${{ matrix.os == 'windows-2019' }}
       uses: actions/upload-artifact@v4
       with:
         name: exports-${{ matrix.architecture }}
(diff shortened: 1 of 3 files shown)


spacex15:
there is a similar problem in planet.cpp

TheGondos:
> there is a similar problem in planet.cpp

Sweet lord, is it the last of the VS2022 issues?

TheGondos:
Changing to draft until everything is sorted out, maybe we should add a VS2022 CI build to prevent future issues

jarmonik:
The latest step might have cause Git CI to crash. Cancelled hung job after several hours of inactivity.

TheGondos:
> The latest step might have caused Git CI to crash. Cancelled job after several hours of inactivity. Status was "Waiting xxx to pickup the job"

Thanks, I saw it was stuck but didn't know how to kill it.
I tried to follow [the documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on), but I now realize that specifying a list limits the matching for a runner instead of making several runs...

dimitry-ishenko:
Maybe drop `windows-2019` altogether, and just leave `windows-2022`? It's only 2024 outside... 😉

jarmonik:
I have very little experience with building the artifacts so can't really advice there. If you think you can direct it to build only 2019 artifact then it's worth trying. Removing the 2019 artifact would reverse the problem, code compiled on 2022 might not run on 2019 but the probability is likely lower. If the first option is giving too much trouble then lets disable VS2019 build, we can restore it if a problem arise.

TheGondos:
Not a fan of switching artifacts to vs2022 right now for the 32bit version, who knows what kind of hidden surprises it involves compatibility-wise.

TheGondos:
Looks like we got a winner, I'll squash everything if you're OK to include this.

jarmonik:
It look good as far as I can tell. 👍

Issue #422, opened 2024-01-21
Keypress detection needed for Lua scripts


ThChicken:
There is no means using Lua to detect keypresses to provide inputs into vessel scripts.

There was a module developed that allowed keypress detection in Lua for Orbiter 2016:

https://www.orbiter-forum.com/threads/trying-to-implement-keypress-detection-in-lua-solved.41000/

but it does not appear to work for OpenOrbiter builds.

schnepe2:
That (C++) code needs to be build against OpenOrbiter ABI to work for OpenOrbiter.
Resulting in an "OpenOrbiter Version" of the `keystate.dll`. The Lua-side should be compatible, I believe.
The source is very simple so it's more a matter of "when" than a matter of "if" 😁

That keystate Lua lib (C++ code to build `keystate.dll`) could be included into OpenOrbiter (provided that the original author likes that).
Please ask the original author whether it's O.K. with him/her. It's not much of a secret code, but still asking is the right way to do it.

ThChicken:
As far as I can tell keystate.dll it was a collaboration between dgatsoulis and martin:

https://www.orbiter-forum.com/threads/trying-to-implement-keypress-detection-in-lua-solved.41000/

TheGondos:
Wouldn't it be cleaner to expose the "clbkConsume*Key" methods to the lua side? (if you're using a ScriptVessel)

ThChicken:
> Wouldn't it be cleaner to expose the "clbkConsume*Key" methods to the lua side? (if you're using a ScriptVessel)

Probably, but right now there isn't anything for Lua in OpenOrbiter at all. Beggars can't be choosers; I'll take whatever we can get done. But yes, it would probably be best to continue trying to get more callbacks exposed so we can make some add-ons without version control becoming a nightmare.

TheGondos:
I'm willing to take a shot at upgrading the ScriptVessel to VESSEL4 and add some callbacks (the easy ones at first) if nobody is working on this already

ThChicken:
Right now Lua has only the following vessel callbacks:

- clbk_setclasscaps
- clbk_postcreation
- clbk_prestep
- clbk_poststep

The clbkConsume*Key callbacks would allow functional add-ons to be made without the need for a special module.

[...]

Issue #421, opened 2024-01-21
Lua Equivalents to clbkSaveState and clbkLoadStateEx Needed


ThChicken:
Lua scripting does not currently have any equivalents to clbkSaveState and clbkLoadStateEx to write or read vessel status information to/from scenario files at the start and end of an Orbiter session.

PR #427, opened 2024-01-26
Update lua capabilities

TheGondos:
This is a WIP for having more functions available in lua through the ScriptVessel.
Should solve issues #421 and #422.


TheGondos:
Coming to a point where it becomes harder to test things (VC stuff at the moment). Is there a toy example that I could use to port to lua somewhere?

TheGondos:
I've been fighting with cross module calls to replicate code like this :
```
void Atlantis::SeparateBoosters (double met)
{
	OBJHANDLE hET = GetDockStatus (GetDockHandle (1));
	if (hET) {
		VESSEL *pET = oapiGetVesselInterface (hET);
		((Atlantis_Tank*)pET)->SeparateSRBs();
	}
}
```
The difficulty is that each vessel lives in its own lua interpreter so you don't have access to code/data from other vessels (even of the same class).
After several false starts, I came up with a solution which forwards calls between interpreters but the syntax is not pretty and I don't know if it is completely supported by lua. It would look like that for the example above :
```
function SeparateBoosters()
  hET = vi:get_dockstatus(vi:get_dockhandle())
  if hET then
    pET = vessel.get_interface(hET)
    pET:import("SeparateSRBs")()
  fi
end

```
It basically works via two functions. lua_import which is added to the ScriptVessel lua instance and returns a closure specific to the method/interpreter
```
// Called from e.g. V:import("SeparateSRBs")
// Returns a closure that will call "SeparateSRBs" inside V's ScriptVessel context
static int lua_import(lua_State* L)
{
	ScriptVessel* sh = lua_toScriptVessel(L, 1);
	if (!sh)
		return 0;

	const char *method = lua_tostring(L, 2);

	// Check if the function exists in the target lua_State
	lua_State* Ltgt = sh->GetState();
	lua_getfield(Ltgt, LUA_GLOBALSINDEX, method);
	if (lua_isfunction(Ltgt, -1) == 0) {
		lua_pop(Ltgt, 1);
		lua_pushnil(L);
		return 1;
	}
	lua_pop(Ltgt, 1);

	// push method name and target lua_State as upvalues for the closure
	lua_pushstring(L, method);
	lua_pushlightuserdata(L, Ltgt);

	// return a closure that will forward the method call to the target lua_State
	lua_pushcclosure(L, lua_crosscall, 2);

	return 1;
}
```
and lua_crosscall which performs the grunt work :
```
// Extract the arguments from the original lua state,
// push them into the target lua state and call the method
// then finally push back the result from the target state to the original state
static int lua_crosscall(lua_State* L) {
	// Recover method name and lua_State for the target interpreter
	const char* method = lua_tostring(L, lua_upvalueindex(1));
	lua_State* Ltgt = (lua_State*)lua_touserdata(L, lua_upvalueindex(2));

	// Save target interpreter stacksize (used later to compute the number of results)
	int stacksize = lua_gettop(Ltgt);

	// Push the function to be called in the target
	lua_getfield(Ltgt, LUA_GLOBALSINDEX, method);

	// Number of arguments pushed in the original lua_State
	int nargs = lua_gettop(L);

	// "Forward" arguments to the target lua_State (Lto)
	// Warning : only basic types are supported (no tables/userdata)
	for (int i = 1; i <= nargs; i++) {
		switch (lua_type(L, i)) {
		case LUA_TNIL:
			lua_pushnil(Ltgt);
			break;
		case LUA_TBOOLEAN:
...
		default:
			// Warning : we push nil for unsupported args to simplify processing (no need to clean up)
			lua_pushnil(Ltgt);
			break;
		}
	}
	// Clean up original state
	lua_pop(L, nargs);

	// Call the function in the target lua state
	lua_pcall(Ltgt, nargs, LUA_MULTRET, 0);
	int nret = lua_gettop(Ltgt) - stacksize; // number of results

	// "Forward" results to the original lua_State *in reverse order*
	for (int i = nret; i >= 1; i--) {
		switch (lua_type(Ltgt, -i)) {
		case LUA_TNIL:
			lua_pushnil(L);
			break;
...
		default:
			lua_pushnil(L);
			break;
		}

	}

	// Cleanup target state
	lua_pop(Ltgt, nret);

	return nret;
}

```
What do you think of this solution? It seems to work, at least with toy examples (calling a luavessel function from a LuaMFD and getting back a couple of values).
There is a limitation where only basic types are supported (don't want to bring in a complete serialization library for this) but I don't think it's that big of an issue.
My concern is if a module function called this way tries to call back into the original module, this could get ugly... Of course it needs a lot more testing (especially to see if it plays nice with coroutines) before I can even put this as a commit in the PR so I'd appreciate if you have any feedback.
Maybe I overthought this and there is a simpler solution?

jarmonik:
I haven't done any work with the LUA before so I am unfamiliar with many of the concepts and issue, but at least @schnepe2 has participated in LUA development.

schnepe2:
Hi,
I am trying to help where I can, but I think Lua ... it's not my strong foot.

Some thoughts however:
It looks like LuAtlantis is just a 1:1 port of Atlantis to Lua. Is this the right way? I would rather think that adding an Lua Interface to Atlantis would make more sense. Any change/impovement in Atlantis you have to do again in LuAtlantis (and vice versa). So from a "do not invent the wheel again" standpoint it seems to be less than perfect.

> Gave it some more thought and ended up detecting global functions automatically so there is no need for the import("xxx") ugliness anymore

I would also recommend not exposing internal states of one Lua instnce to another, but I cannot really tell why 😉

TheGondos:
> Some thoughts however: It looks like LuAtlantis is just a 1:1 port of Atlantis to Lua. Is this the right way? I would rather think that adding an Lua Interface to Atlantis would make more sense. Any change/impovement in Atlantis you have to do again in LuAtlantis (and vice versa). So from a "do not invent the wheel again" standpoint it seems to be less than perfect.

The goal is to improve Lua support to the point where it can be a viable alternative to C++ modules. Some kind of prototype vessel is required to cover the added Lua functionalities.
I chose to port the Atlantis 1 to 1 because :
- I'm not a modder and acquiring the skills to create a complex vessel from scratch would take too much time for the task at hand (not that I would mind being able to do that)
- It's one of the few existing vessels under the same MIT license as the Orbiter project (obviously since it's part of it) so I can put it in the repo as an example of how to implement a relatively complex vessel in Lua
- It provides a reference to check against when there are odd behaviours with the Lua version

> I would also recommend not exposing internal states of one Lua instnce to another, but I cannot really tell why 😉

Don't worry, only global functions are accessible. You just have to declare private stuff as local for it to be hidden.

It's my first experience with Lua so I'm pretty much playing it by ear at the moment, don't hesitate to tell if I do something stupid^^

ThChicken:
Saw that clbkSaveState/clbkLoadStateEx capabilities were added, but there were some test problems? Are these available in a release build, or are all of these methods going to be released together at a later time? I'm just unfamiliar with GitHub and don't know what capabilities are publicly available. Thanks for working on these Lua capabilities!

schnepe2:
> Saw that clbkSaveState/clbkLoadStateEx capabilities were added, but there were some test problems?

Just **automatic** CI[*] testing was not working. The functionality can only tested locally.
Nevertheless the functions are available in master as they heve been merged.

> Are these available in a release build, or are all of these methods going to be released together at a later time?

Yes they are available, as long as the builds were created from master (which they are). Those builds contain all the changes that were merged/pulled into the master - so "latest release" contains all the "latest changes" (including latest bugs 😉 )

>  I'm just unfamiliar with GitHub and don't know what capabilities are publicly available. Thanks for working on these Lua capabilities!

No problem, the distributed development via github is another level of complexity. But after a few times (and stupid errors on my side) I think I've got a hang on it now.

[*] Continous Integration

schnepe2:
Oh sorry, I was thinking of another PR... The clbkSaveState/clbkLoadStateEx functions have only been added to this PR.
So they are only available in builds from this "branch". As it is currently only on the creators repository it is NOT included in the master builds (yet).

The builds you see in this PR (Those of the checks) however are not stored, so you cannot download them (as far as I know).

Sorry for the confusion.

ThChicken:
> Oh sorry, I was thinking of another PR... The clbkSaveState/clbkLoadStateEx functions have only been added to this PR. So they are only available in builds from this "branch". As it is currently only on the creators repository it is NOT included in the master builds (yet).
> 
> The builds you see in this PR (Those of the checks) however are not stored, so you cannot download them (as far as I know).
> 
> Sorry for the confusion.

No worries. I've been updating my Lua vessel code in anticipation of these features. Looking forward to seeing these in the master build.

TheGondos:
> > Oh sorry, I was thinking of another PR... The clbkSaveState/clbkLoadStateEx functions have only been added to this PR. So they are only available in builds from this "branch". As it is currently only on the creators repository it is NOT included in the master builds (yet).
> > The builds you see in this PR (Those of the checks) however are not stored, so you cannot download them (as far as I know).
> > Sorry for the confusion.
> 
> No worries. I've been updating my Lua vessel code in anticipation of these features. Looking forward to seeing these in the master build.

I hope I'll be able to change the PR to ready status next weekend so it can be one step closer to general availability.
Sadly if you want to play with it you must compile it from this branch.

TheGondos:
The MFD work is pushed. Since the registered MFDs are Lua objects (tables), I could not reuse the ScriptMFD since it expects a script file and uses global functions. I had to create another "VesselMFD" container to encapsulate the C++ MFD along with the Lua class.

TheGondos:
The initial goal is pretty much done, there is still some missing API for the 2D panel views but I think it would best be done in another PR, this one is big enough already.
I still have to complete the vessel/oapi documentation, since it's really tedious, I'd like to have some feedback before doing it. It's not something I want to do twice...

ThChicken:
> The initial goal is pretty much done, there is still some missing API for the 2D panel views but I think it would best be done in another PR, this one is big enough already. I still have to complete the vessel/oapi documentation, since it's really tedious, I'd like to have some feedback before doing it. It's not something I want to do twice...

I'm not super familiar with github or how the code is managed, just a highly interested user. But if there is something I can do to facilitate the review and documentation I'm happy to help.

TheGondos:
> > The initial goal is pretty much done, there is still some missing API for the 2D panel views but I think it would best be done in another PR, this one is big enough already. I still have to complete the vessel/oapi documentation, since it's really tedious, I'd like to have some feedback before doing it. It's not something I want to do twice...
> 
> I'm not super familiar with github or how the code is managed, just a highly interested user. But if there is something I can do to facilitate the review and documentation I'm happy to help.

This is quite a paradox, the idea is to be able to work without a C++ environment yet you need to compile the branch to be able to test it...
Maybe if you can grab the latest OpenOrbiter, I can attach the DLLs (LuaInline, LuaInterpreter and ScriptVessel) ?

ThChicken:
> > > The initial goal is pretty much done, there is still some missing API for the 2D panel views but I think it would best be done in another PR, this one is big enough already. I still have to complete the vessel/oapi documentation, since it's really tedious, I'd like to have some feedback before doing it. It's not something I want to do twice...
> > 
> > 
> > I'm not super familiar with github or how the code is managed, just a highly interested user. But if there is something I can do to facilitate the review and documentation I'm happy to help.
> 
> This is quite a paradox, the idea is to be able to work without a C++ environment yet you need to compile the branch to be able to test it... Maybe if you can grab the latest OpenOrbiter, I can attach the DLLs (LuaInline, LuaInterpreter and ScriptVessel) ?

I can do that. I have the x86 602931718 (February 1) build which I believe is the latest version.

But you have done a lot of code revision - what exactly would I be reviewing/testing? What do you need me to do?

ThChicken:
OK. I installed the DLLs and clbk_savestate works correctly for my Mirage2000 vessel.

There seems to be some problem with clbk_loadstateex. I added the function but left it empty. I assume the second argument must be the vessel instance handle (vi in my code)?
`
function clbk_loadstateex(scn, vi)
end
`
On start Orbiter terminates with a critical error dialog, and I get the attached Orbiter.log.
[Orbiter.log](https://github.com/orbitersim/orbiter/files/14393056/Orbiter.log)

It looks like LuaInLine is trying to read a filename out of the scenario file (???). Not sure what to make of this.

I can work on converting over the keypress functions once we get this sorted.

ThChicken:
I am also looking at  vi:create_variabledragelement to replace the add_force calls for landing gear drag. Looking at the Atlantis code it seems that there are only two arguments (the force magnitude and the vector of the point of attack). In the C++ API reference this method has a third argument of the control parameter, which I would put in as the animation state of the gear. I'm not sure how this can be a variable component without that.

TheGondos:
> OK. I installed the DLLs and clbk_savestate works correctly for my Mirage2000 vessel.

Nice!
 
> There seems to be some problem with clbk_loadstateex. I added the function but left it empty. I assume the second argument must be the vessel instance handle (vi in my code)? `function clbk_loadstateex(scn, vi) end` On start Orbiter terminates with a critical error dialog, and I get the attached Orbiter.log. [Orbiter.log](https://github.com/orbitersim/orbiter/files/14393056/Orbiter.log)
> 
> It looks like LuaInLine is trying to read a filename out of the scenario file (???). Not sure what to make of this.
> 

If you implement the clbk_loadstateex callback, then you absolutely need to forward the unrecognized lines to the default scenario handler (this is the same in C++) or the Orbiter parsing logic will not correctly detect the end of the vessel declaration and will consider the second line as a new ship declaration (hence the "No vessel class configuration file found for: STATUS Landed Earth" error).
For example :
```
function clbk_loadstateex(scn, vs2)
     for line in scenario_lines(scn) do
		if scenario_line_match(line, "CONFIGURATION %d", match) then
			status = match.res[1]
		elseif scenario_line_match(line, "GEAR %d %f", match) then
			gear_state.status = match.res[1] + 1
			gear_state.proc = match.res[2]
                else
                        vi:parse_scenario_line_ex(line, vs2)  -- unrecognised option - pass to Orbiter's generic parser
                end
     end
end

```
Also there is already a _vi_ global variable populated by default by the ScriptVessel that corresponds to the current vessel object. I would advise not using _vi_ as a parameter or variable in your code as it will probably lead to problems.

TheGondos:
> I am also looking at vi:create_variabledragelement to replace the add_force calls for landing gear drag. Looking at the Atlantis code it seems that there are only two arguments (the force magnitude and the vector of the point of attack). In the C++ API reference this method has a third argument of the control parameter, which I would put in as the animation state of the gear. I'm not sure how this can be a variable component without that.
> 
> EDIT: Nevermind, I see you have a separate set command to set the drag level. It appears to work correctly, but I don't know if that will be a point of confusion for developers trying to use the 3-argument form of CreateVariableDrag.

Yes, I'm not really happy with that solution but since there are no references in Lua, I seemed to me to be the best compromise. I'll stress that point in the documentation

ThChicken:
> Yes, I'm not really happy with that solution but since there are no references in Lua, it seemed to me to be the best compromise. I'll stress that point in the documentation

OK. I'm unfamiliar with the inner workings of the code and I don't understand what you mean by "no references" being a problem. Is that something you can explain in plain language? Just curious.

ThChicken:
> Edit: if you want to use the scenario_lines/scenario_line_match functions, you need to put [this oapi_init.lua](https://raw.githubusercontent.com/TheGondos/orbiter/9fb97022aa50384c1955ac52040afaccc86de18a/Script/oapi_init.lua) file in your Script directory

OK. But I should still be able to parse the scn file with just Lua syntax, yes?

> Edit2 : I missread your comment about the second argument. No, it's not the vessel interface, it's an opaque handle to a VESSELSTATUS2 structure that you need to provide to the parse_scenario_line_ex method.

OK. Again, the terminology is over my head. Is the handle vs already available or does it need to be referenced/defined in some way?

ThChicken:
> > OK. I installed the DLLs and clbk_savestate works correctly for my Mirage2000 vessel.
> 
> Nice!

Actually, it mostly works in that it can write out numerical animation statuses and states, but it isn't writing out strings:
`
function clbk_savestate(scn)
.
.
.
oapi.writescenario_string(scn, "afterburner", afterburner) -- this is ignored for some reason? afterburner boolean does exist
.
.
.
end
`

TheGondos:
> 
> Actually, it mostly works in that it can write out numerical animation statuses and states, but it isn't writing out strings:
> 
> oapi.writescenario_string(scn, "afterburner", afterburner) -- is ignored

Are you sure your afterburner variable is a string and not a boolean? If not you must convert it with tostring(afterburner)

> OK. But I should still be able to parse the scn file with just Lua syntax, yes?
> 
Yes

> OK. Again, the terminology is over my head. Is the handle vs already available or does it need to be referenced/defined in some way?

The only hint I have is from the C++ headers :
```
//@{
	/**
	 * \brief Pass a line read from a scenario file to Orbiter for default processing.
	 * \param line line to be interpreted
	 * \param status status parameters (points to a VESSELSTATUSx variable).
	 * \note This function should be used within the body of \ref VESSEL2::clbkLoadStateEx.
	 * \note The parser clbkLoadStateEx should forward all lines not recognised
	 *   by the module to Orbiter via ParseScenarioLineEx to allow processing of
	 *   standard vessel settings.
	 * \note clbkLoadStateEx currently provides a VESSELSTATUS2 status definition.
	 *   This may change in future versions, so status should not be used within
	 *   clbkLoadStateEx other than passing it to ParseScenarioLineEx.
	 * \sa VESSEL2::clbkLoadStateEx
	 */
	void ParseScenarioLineEx (char *line, void *status) const;
	//@}
```
So you must take the second parameter from clbk_loadstateex, whatever the name you give it, and pass it to the parse_scenario_line_ex function, like in the example I showed.

> OK. I'm unfamiliar with the inner workings of the code and I don't understand what you mean by "no references" being a problem. Is that something you can explain in plain language? Just curious.

I'm not sure but I can try ^^
CreateVariableDragElement in C++ takes a pointer to an "external drag control parameter" that is usually the animation state of a moving control surface, so that when the animation changes the drag function automatically uses the new value.
There is no valid way from the C++ side to get a pointer to a Lua variable so we must use another method.

If you want the details, the create_variabledragelement implementation allocates a memory block that will be used to store the "external drag control parameter", and pass it both to CreateVariableDragElement and as a return value to the Lua script. Then from Lua when you use the "set" command, it updates the memory block and the drag function performs its calculation with the correct value.

ThChicken:
> Are you sure your afterburner variable is a string and not a boolean? If not you must convert it with tostring(afterburner)
Oh bother, of course.

I'm actually having troubles with the oapi_init.lua file you provided - Orbiter crashes on start and I don't believe it is anything in my script. I'll try to copy and paste it again.

TheGondos:
I placed every ScriptVessel callbacks behind lua_pcall calls to intercept errors and prevent CTD. The last error should pop at the bottom of the screen. Here are the DLLs if you want to test against OpenOrbiter : [luavessel2.zip](https://github.com/orbitersim/orbiter/files/14412252/luavessel2.zip)

TheGondos:
Opps, this does not play well with CI tests...

ThChicken:
OK, I was suffering from a self-inflicted code issue. To summarize my findings:

1. clbk_savestate(scn) works. Booleans must be converted to strings using the Lua tostring() function of course.
2. clbk_loadstateex(scn, vs2) works and there is no problem with the oapi_init.lua file you provided to implement the functions. Booleans can only be read in as strings and must  be converted back to into booleans "if boolean == "true" then boolean = true", etc.. This is known Lua behavior, but probably should be mentioned somewhere in the documentation.

So I am able to open a scenario with my Script vessel, fly it and change the animation and other states, exit Orbiter, and resume with the (Current state).scn file and pick up the simulation right where I left off. Progress! Well done!

I have seen your posts regarding LuaL_error and the annotations and it looks very promising. I will need to look into what you are doing there in more detail as I tend to make seemingly innocuous changes in my code which actually cause some conflict and the script simply starts crashing, and I have to do start commenting code until I isolate the issue. I can do it but it is aggravating. If you have figured out some way to get some more verbosity about errors to us that is great news.

I will now start looking at detecting keypress events with clbk_consumebufferedkey and see if I can get that working with my add-on.

TheGondos:
>I confess I don't quite know how to test clbk_consumebufferedkey(kstate) and I don't really know what that does relative to clbk_consumedirectkey(kstate).

clbk_consumedirectkey is called every frame with the whole keyboard state.
clbk_consumedirectkey is called only when you press/release a key. It provides the key, event and the whole keyboard state so that you can test for modifiers (ctrl, alt, shift...)

> I have seen your posts regarding LuaL_error and the annotations and it looks very promising. I will need to look into what you are doing there in more detail as I tend to make seemingly innocuous changes in my code which actually cause some conflict and the script simply starts crashing, and I have to do start commenting code until I isolate the issue. I can do it but it is aggravating. If you have figured out some way to get some more verbosity about errors to us that is great news.

Did you test with the DLLs from luavessel2.zip posted above? It includes the onscreen error reporting.

Thanks for your effort :-)

ThChicken:
> clbk_consumedirectkey is called every frame with the whole keyboard state. clbk_consumedirectkey is called only when you press/release a key. It provides the key, event and the whole keyboard state so that you can test for modifiers (ctrl, alt, shift...)

You refer to consumedirectkey twice - I think one of these was meant to be consumebufferedkey? As I mentioned, clbk_consumedirectkey works fine, but I don't know what is different about clbk_consumebufferedkey or if I can test it with my current vessel script.
 
> Did you test with the DLLs from luavessel2.zip posted above? It includes the onscreen error reporting.

I just downloaded and replaced those DLLs. I am getting a bunch of errors related to code in separate Lua modules that complain that thruster group names are undefined/nil. Indeed, they are not defined in the module, but they are defined in the main Lua script that calls those modules. I am not sure if this is actually an error or an artifact of my using Lua module files.

>Thanks for your effort :-)

Thanks for all of yours! We should be able to make some real add-ons for Orbiter now with Lua script thanks to your efforts.

TheGondos:
> You refer to consumedirectkey twice - I think one of these was meant to be consumebufferedkey? As I mentioned, clbk_consumedirectkey works fine, but I don't know what is different about clbk_consumebufferedkey or if I can test it with my current vessel script.

Oops, the clbk_consumebufferedkey is call on key press. You could for example set your afterburner variable to true on a keydown event and false on keyup, no need to check everyframe.

> I just downloaded and replaced those DLLs. I am getting a bunch of errors related to code in separate Lua modules that complain that thruster group names are undefined/nil. Indeed, they are not defined in the module, but they are defined in the main Lua script that calls those modules. I am not sure if this is actually an error or an artifact of my using Lua module files.
 
If errors are showing then your code is not executed completely. Are you defining the group names as local or global variables? I don't think a module's local variables are visible from it's "required" modules.

ThChicken:
> You could for example set your afterburner variable to true on a keydown event and false on keyup, no need to check every frame.

I might want to switch my implementation from clbk_consumedirectkey to clbk_consumebufferedkey then. I noticed that I get strange behavior if I hold the key down.
 
> If errors are showing then your code is not executed completely. Are you defining the group names as local or global variables? I don't think a module's local variables are visible from it's "required" modules.

I have my air-breathing engine code in that module.The thruster group names are global. The code runs correctly. I wouldn't have any thrust if the jet engine module wasn't working properly. It requires a main and a dummy thruster to take the throttle input and pass it into the engine module. Not sure how to code the module to avoid this. I suppose I can just put the functions into my main code but it rather clutters things up.

TheGondos:
> I have my air-breathing engine code in that module.The thruster group names are global. The code runs correctly. I wouldn't have any thrust if the jet engine module wasn't working properly. It requires a main and a dummy thruster to take the throttle input and pass it into the engine module. Not sure how to code the module to avoid this. I suppose I can just put the functions into my main code but it rather clutters things up.

Definitely use a module. You can PM me your code on OF if you want, I can take a look at what's going on.

ThChicken:
> You could for example set your afterburner variable to true on a keydown event and false on keyup, no need to check every frame.

I suppose I need some clarification on the terminology 'keydown' and 'keyup' - I am assuming that 'keydown' is when the key is pressed, and 'keyup' is when it is released? So it basically makes the key act like a momentary switch?

> > I have my air-breathing engine code in that module.The thruster group names are global. The code runs correctly. I wouldn't have any thrust if the jet engine module wasn't working properly. It requires a main and a dummy thruster to take the throttle input and pass it into the engine module. Not sure how to code the module to avoid this. I suppose I can just put the functions into my main code but it rather clutters things up.

> Definitely use a module. You can PM me your code on OF if you want, I can take a look at what's going on.

I may do this, but with your recent code improvements my vessel code is getting simpler, and I am starting to see some inefficiencies in my code. I think I could pass thruster names and variables without using globals, but I'd have to rewrite a lot of code. Let me clean up my code as best as I am able, and then maybe I'll take you up on your offer to look at it.

ThChicken:
Ok, I see it now. clbk_consumebufferedkey works nicely. clbk_consumedirectkey seems to be sensitive to how long one holds the key down, which I think is expected behavior. clbk_consumebufferedkey toggles the animations more smoothly and consistently.

ThChicken:
I have encountered a problem with the scenario_line_match functions - I thought they were reading my variables from the scenario file but apparently they were not and my initialization values were being used. 

If I start a simulation and exit, the variables are written correctly to the scenario file. But if I restart with that scenario file, it won't read the values. The scenario file seems to be 100% correct, so I am rather confused as to where the problem is. I am considering trying to parse the scenario file using Lua methods.

Here is the code and the scenario file:
```
function clbk_loadstateex(scn, vs2)   
    match = {}
    for line in scenario_lines(scn) do
        if scenario_line_match(line, "afterburner %s", match) then
            if match.res[1] == "true" then
                afterburner = true
            elseif match.res[1] == "false" then
                afterburner = false
            end
        elseif scenario_line_match(line, "gear_anim_state %d", match) then
            gear_anim_state = match.res[1]
        elseif scenario_line_match(line, "gear_anim_status %d", match) then
            gear_anim_status = match.res[1]
        elseif scenario_line_match(line, "canopy_anim_state %d", match) then
            canopy_anim_state = match.res[1]
        elseif scenario_line_match(line, "canopy_anim_status %d", match) then
            canopy_anim_status = match.res[1]
        else
            vi:parse_scenario_line_ex(line, vs2)  -- unrecognised option - pass to Orbiter's generic parser
        end
    end
end
```
```
BEGIN_URLDESC
CurrentState_img
END_URLDESC

BEGIN_ENVIRONMENT
  System Sol
  Date MJD 51983.4234207768
  Help CurrentState_img
END_ENVIRONMENT

BEGIN_FOCUS
  Ship Mirage2000Script
END_FOCUS

BEGIN_CAMERA
  TARGET Mirage2000Script
  MODE Extern
  POS 4.000000 0.000000 0.000000
  TRACKMODE TargetRelative
  FOV 40.00
END_CAMERA

BEGIN_HUD
  TYPE Surface
END_HUD

BEGIN_MFD Left
  TYPE Surface
  SPDMODE 1
END_MFD

BEGIN_SHIPS
Mirage2000Script:Mirage2000Script
  STATUS Orbiting Earth
  RPOS 4009379.065 4951024.800 218043.284
  RVEL 350.1001 -236.8856 319.5351
  AROT -65.376 -60.681 30.346
  VROT 0.1196 0.9666 0.0214
  RCSMODE 0
  AFCMODE 7
  PRPLEVEL 0:0.100000
  THLEVEL 0:0.991926 1:1.000000 2:0.000174
  NAVFREQ 94 524
  afterburner true
  gear_anim_state 1.0
  gear_anim_status 0
  canopy_anim_state 0.0
  canopy_anim_status 0
END
END_SHIPS
```

TheGondos:
> If I start a simulation and exit, the variables are written correctly to the scenario file. But if I restart with that scenario file, it won't read the values. The scenario file seems to be 100% correct, so I am rather confused as to where the problem is. I am considering trying to parse the scenario file using Lua methods.

Is it only the states that are not loading properly?
%d matches only integers, you need to use %f for floats :

`elseif scenario_line_match(line, "gear_anim_state %f", match) then`

ThChicken:
> Is it only the states that are not loading properly? %d matches only integers, you need to use %f for floats 

I corrected them to %f for the floats but it doesn't change the behavior. I can initiate a simulation, fly around a bit, and then exit. The variables are written to the scenario file correctly. When I go to restart the scenario, it crashes. Any scenario that has these variables assigned to the vessel will crash on open. I can't really see anything with the ScriptMFD module that annotates the errors as it is buried in the warnings about uninitialized thruster names in my Lua modules.

> I will change the %b option to accept true/false strings so you can get rid of this logic :
> 
> ```
>             if match.res[1] == "true" then
>                 afterburner = true
>             elseif match.res[1] == "false" then
>                 afterburner = false
>             end
> ```

Thanks!

ThChicken:
I found my self-inflicted code issues. As far as I can see keypress detection and reading/writing vessel status to/from the scenario files all works flawlessly. I'm happy to call issues #421 and #422 resolved whenever these changes get released. 

I'm flying my all-Lua vessel add-on with my all-Lua autopilot MFD on a great circle mission, and I can save it and resume it without any problems! Well done! Thanks for all of your hard efforts!

If there is anything else I can test for you I'm happy to help out.

jarmonik:
Is this ready for Merge ?

TheGondos:
> Is this ready for Merge ?

Can we wait till sunday?
I still got some work that I'd like to push but I don't know if it'll be done by then.
If it's not ready by sunday, we can do without it for now (it's mostly panel2d/VC stuff)
I'll squash everything when it's OK to be merged, 30+ commits is a bit overkill.

jarmonik:
Yes of course. I noticed PR is no longer 'Draft' thus the question. Take all the time you need no rush. Give it 'Ready for Merge' when ready.

TheGondos:
Took longer than expected because I had to set up a test vessel for the virtual cockpit/2D panel work. It's a rewrite of the DeltaGlider in Lua, quite incomplete at the moment but good enough to test the new API.
There is a now also a StrictMode option available for ScriptVessel to help in debugging.

TheGondos:
I squashed the whole 44 commits into 2 to clean up the history. One for the C++ part, and one for the Lua vessels provided as examples in the SDK.
I'm not sure how to handle the documentation with the current Latex work being done.
Apart from that, it should be good to merge.

TheGondos:
I see there was some docking update in the main branch, let me sync this one to take it into account

TheGondos:
I think I'm finally done with this one. You can merge it if you're OK with it.

PR #428, opened 2024-01-29
Bug fix patch

jarmonik:
Number of bug fixes:
- Vessel sunlight bug near sun
- Fixed flickering "doors" in DG in planetarium mode.
- Max mesh resolution set to 32 to avoid out of video memory
- Fixed TerrainToolKit multi-byte text display issue.
- Added warning for missing planetary textures. Popup window if gravity-ref is missing textures.
- Fixed sun-glare visibility near sun.  
- Fixed sunlight occlusion by body other than the closest one
- Fixed a problem of moon being lit in a shadow of a planet by adding (accurate) support for eclipses.
- Made a tile name appear on red in TerrainToolKit if the tile doesn't exists.


n7275:
I ask mostly out of curiosity, but what was the cause of this? I wasn't able to exactly follow in the code.

jarmonik:
There were multiples issues. First of all a star visual was incorrectly casted to vPlanet. (It kinda odd that it didn't crash right away). As a result of that the sun light color become NaN. I still need to do some testing by taking the DG in various other places.

n7275:
> There were multiples issues. First of all a star visual was incorrectly casted to vPlanet. (It kinda odd that it didn't crash right away). As a result of that the sun light color become NaN. I still need to do some testing by taking the DG in various other places.

Oh wow. yeah, that would do it.
If I have some time tonight, I'm more than happy to build this branch, and go on a little "grand tour" to test it out.

jarmonik:
There are still at-least two bug those are both related to a bad surface reference. Strongest gravity source is not always the best reference and using the nearest body also causes problems.

n7275:
> There are still at-least two bug those are both related to a bad surface reference. Strongest gravity source is not always the best reference and using the nearest body also causes problems.

Would something like "object with largest angular size" work?

jarmonik:
> Would something like "object with largest angular size" work?
Probably not. One would need to be pretty close to Phobos until it's larger than Mars. Also the SOI of Io is very small inside Jupiter's gravity.

jarmonik:
I changed the near body approximation to max of _mass^0.33 / distance^2_

PR #438, opened 2024-03-04
Conversion of existing documentation to LaTeX sources

GLS-SSV:


Doc/Orbiter Developer Manual/PUBLISH.tex
```diff
@@ -0,0 +1,44 @@
+\documentclass[Orbiter Developer Manual.tex]{subfiles} 
+\begin{document}
+
+\section{Publishing addons}
+Now that you have created and tested your new ship, you want to share it with the rest of the Orbiter community. Here is how to do it:
+
```
TheGondos, line 6: Maybe we should add some words about licencing in this section ?

GLS-SSV: Makes sense. Added a "TODO" in that section to indicate the need for this (this PR is for LaTeX conversion work, the next PR will work the updating and checking of the content).

jarmonik:
I tried to build the documentation but ran into some problems. The used font "NotoSansMono" doesn't seem to work. I found one from the internet but it refuses to install. Is there something special about the font that makes it better ? Or would it be practical to change to a more common font ?

I renamed the font from common_definitions.tex to "Calibri" and I was able to build Technical Reference and Developer Manual but Orbiter User Manual fails to build with following error:

> !pdfTeX error: pdflatex.exe (file fstr8x): Font fstr8x at 624 not found
> ==> Fatal error occurred, no output PDF file produced!

jarmonik:
Still no luck getting this to work. Looks like the "Noto" font is a part of MikTex installation and located in /MikTex/fonts/truetype/google/noto/  But why doesn't it use it ?  I am trying to build the documentation with TeXnicCenter.

Is anyone else able to edit and review the documentation locally ?

PR #446, opened 2024-03-16
Memory recycle experiment

jarmonik:
Reuse memory instead constant release and reallocation.


n7275 (approved):

OVP/D3D9Client/D3D9Catalog.h
```diff
@@ -82,27 +85,270 @@ class Memgr
 
 	T* New(DWORD size)
 	{
+		mm.lock();
```
dimitry-ishenko, line 90: FYI you can use `std::lock_guard` instead of manually locking and unlocking.

PR #448, opened 2024-03-21
Install of SDK samples and LUA samples

jarmonik:


jarmonik:
This would need double checking that everything is ok

n7275 (approved):

PR #450, opened 2024-03-22
Fix for custom problem

jarmonik:
- Fixed the custom camera problem (tiles not loading)
- Fixed debug assert on camera setup.
- Added launchpad option for max tile count.
- Fixed a problem causing black caps between tiles. (may still occur if tile lvl difference too high)
- Fixed the camera distance reset on view mode change.
- Fixed cache performance problem (Thanks to patio)
- Fixed vessel surface flicked caused by vessel markers.


n7275 (approved):

jarmonik:
There are still a few problems requiring a fix. I'll inform when ready for review.

jarmonik:
This should be ready for review and testing. Hopefully last bug fix PR for Orbiter 2024

n7275 (approved):

PR #451, opened 2024-03-27
[Lua]Fix Lua DG inputs + errors when loading/saving scenario

TheGondos:
Fixes for the Lua DG SDK :
-mismatch between boolean and integer in key handling
-error when saving scenario with non default instrument light
-error when loading scenario with non default cockpit lights


TheGondos:
Ready for merge

PR #455, opened 2024-04-05
Bug fixes and Improvements

jarmonik:
- Fixed Ctrl+Shift+Arrow crash


n7275:
Does this need more work to be merge-able? I don't quite understand what was causing the crash that this fixes.

jarmonik:
> Does this need more work to be merge-able? I don't quite understand what was causing the crash that this fixes.
Not really. I am looking into the micro-texture problems and hoping to bundle with this PR. The crash is related to gcGUI which must be enable to produce the crash.

PR #456, opened 2024-04-07
[Lua]Add support for oapiOpenInputBoxEx

TheGondos:
This PR adds support for the oapiOpenInputBoxEx function in ScriptVessels/ScriptMFDs
Differences with the C++ version :
-callbacks only take 2 arguments (str and usrdata) since the first "void *id" is useless
-cancel callback does not need to return a bool since it's ignored by the core anyway

### DIFF
--- a/Script/strict.lua
+++ b/Script/strict.lua
@@ -42,3 +42,5 @@ end
 function strictmode_add_global(name)
 	mt.__declared[name] = true
 end
+-- Global used for aborting coroutines
+wait_exit = nil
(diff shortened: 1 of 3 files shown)


TheGondos:
What's the strategy for the Lua documentation? I don't like adding new APIs without at least minimal documentation...

jarmonik:
There is no strategy for LUA documentation as far as I can tell. So, would you have some recommendations ?
Would it be possible to exploit doxygen and modified versions of OrbiterAPI.h and VesselAPI.h for documentation ?

TheGondos:
OK, I was not sure if the API doc was handled in the LaTeX work or not. I'll try to see what I can do with the ldoc stuff lying around then.
I won't do this in this PR so it's ready to merge

PR #461, opened 2024-05-31
Case sensitive filesystem support

mdluhosch:
Hi, 

I am currently playing around with this repo under Linux.  One issue I had is that linux file systems are case sensitive. I thought maybe you want to have the "correct" names also in the main repo.
Also on some spots the \<algorithm\> header was missing (on linux the std::find does not work without it)
The change in MfdAlign.cpp seems to be needed because the winegcc compiler is a bit more strict. 

Kind Regards
Michael

### DIFF
--- a/Src/Orbiter/Astro.h
+++ b/Src/Orbiter/Astro.h
@@ -8,7 +8,7 @@
 #define __ASTRO_H
 
 #include <time.h>
-#include "vecmat.h"
+#include "Vecmat.h"
 #include "OrbiterAPI.h"
 
 const double iAU = 1.0/AU;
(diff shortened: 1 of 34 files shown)


TheGondos:
I'm curious why you did not have problems with Orbitersdk.h/orbitersdk.h?

mdluhosch:
I did not compile everything so far. Currently have some trouble with wine on my distro. So probably I did not yet stumble upon source files with the "incorrect" orbitersdk.h

TheGondos:
Maybe someone could try [enabling case sensitivity](https://www.windowscentral.com/how-enable-ntfs-treat-folders-case-sensitive-windows-10) on a Windows box and see what happens. It would be easier to do a complete check this way.

Good luck with your plan. I gave up compiling with winegcc, it was easier to port the whole thing to linux^^

jarmonik:
What is a status of this ?  Obsolete ?

PR #475, opened 2024-07-27
Perform pre-check before running nightly

DarkWanderer:
Purpose: check that it is worth running another instance of nightly build


dimitry-ishenko:
Thanks for doing that

PR #478, opened 2024-08-13
Finishing touches

jarmonik:
- Fixed Neptune lighting issue. Further testing needed.
- Disabled MSAA when using mesh flag 0x20 to avoid edge glow.
- Enabled full range of charters for text output.


n7275:
Awesome. I'll give it a test as soon as I can.
This should, in theory, close 466, and close 477

n7275:
The lighting issue appears to be solved.

PR #483, opened 2024-08-21
[Lua]Add radiation pressure callback and oapiSetMeshProperty

TheGondos:
This PR adds support for the clbkGetRadiationForce callback in Lua.
I started to port the SolarSail vessel but its visual is broken and only works with MOGE so I'll leave it as it is for the time being.
As a result the oapiSetMeshProperty has been added too since the SolarSail was making use of it.


TheGondos:
Added SolarSail in Lua. Can be a bit sluggish because of the way the sail is animated but should be good enough as an example.

PR #484, opened 2024-08-22
Use WinGet instead of Chocolatey

DarkWanderer:
Attempt to reduce documents prerequisites installation time


DarkWanderer:
`winget` is not fit for this purpose

PR #488, opened 2024-08-31
Fix SolarSail dynamic geometry with DX9

TheGondos:
This PR fixes the SolarSail vessel. It was using the old oapiMeshGroup API to patch vertex positions so that it only worked with MOGE.
Change it to use oapiEditMeshGroup instead.


jarmonik:
Is this ready for merge or additional testing required ?

TheGondos:
Would be nice to have some more feedback but I doubt we'll get some. We might as well merge it, the impact is limited to this one vessel.

n7275 (approved):

PR #489, opened 2024-08-31
[DX9]Fix oapiSetMaterialEx

TheGondos:
@jarmonik can confirm but it looks like there's been some kind of naming confusion between clbkSetMaterialEx and clbkSetMeshMaterialEx.
oapiSetMaterialEx in the core calls clbkSetMeshMaterialEx in the GC.
The D3D9Client implements a clbkSetMaterialEx method which is used by the gcCore API but it will not match when calling oapiSetMaterialEx and the default implementation will be used (i.e. "return 2;").
I modified the D3D9Client to implement the correct naming but it could also be done by changing the GC interface in the core.
Don't know what's best between naming consistency in the core (clbkSetMeshMaterial, clbkMeshMaterial and clbkMeshMaterialEx also exist) or the GC ; tell me if you think it should be done the other way around.

### DIFF
--- a/OVP/D3D9Client/D3D9Client.cpp
+++ b/OVP/D3D9Client/D3D9Client.cpp
@@ -1507,7 +1507,7 @@ int D3D9Client::clbkMeshMaterial (DEVMESHHANDLE hMesh, DWORD matidx, MATERIAL *m
 
 // ==============================================================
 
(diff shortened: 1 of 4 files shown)


jarmonik:
Looks good to me. Thanks :)

PR #490, opened 2024-09-03
fix: tab out/scroll mouse permanently disappearing bug

ZigTag:
I was talking to some of the community members in the NASSP discord, and they informed me that the disappearing mouse bug that would happen when tabbing out and zooming while scrolling over a switch while being in the right click (RotationMode), was not an isolated bug. So I decided to take it upon myself to fix it.

The problem is that the reference counter on the ShowCursor function would increment too high due to loosing focus, so it checks that it is below a certain threshold and tries to rectify it by calling the function again.

### DIFF
--- a/Src/Orbiter/Orbiter.cpp
+++ b/Src/Orbiter/Orbiter.cpp
@@ -99,7 +99,7 @@ const double    MinWarpLimit     = 0.1;  // make variable
 const double    MaxWarpLimit     = 1e5;  // make variable
 DWORD           g_qsaveid        = 0;
 DWORD           g_customcmdid    = 0;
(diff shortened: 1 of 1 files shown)


ThatRedMelon:
This has been a highly annoying bug for many years, I would highly recommend including this PR in the Orbiter2024 release.

TheGondos:
Did several trials of alt-tabbing while rotating the view with the mouse. The cursor always reappeared when right-clicking to cancel rotation mode after getting the focus back.
Without the patch, the cursor bug appeared quickly doing the same scenario.

ZigTag:
I'mma just move everything to #491

PR #491, opened 2024-09-05
fix: rotationmode now takes mouse priority over panel buttons + alt-tab mouse disappearing fix

ZigTag:
This PR patches the bug where the mouse could disappear when alt tabbing from the game, this was an almost game-breaking bug where you would have to restart your entire sim in order to fix the bug. The mouse now has a global hide counter to check for if it is already hidden, and prevents it from being hidden twice, this makes it so it can never be permanently hidden, as ShowCursor(true) only increments the counter by 1, whereas before it'd get stuck below 0 and therefore stay permanently hidden. 

After patching the alt-tab bug, I noticed a bug in NASSP where when moving and scrolling the mouse, it would activate unintended buttons when trying to change the FOV. Rather than changing the FOV it would perform the "right click" action on any button while scrolling. I do not believe this is intended behavior, so I created a shim that activates when the camera rotation mode activates which prevents it from doing anything other than mouse movement, and camera interaction. This shouldn't affect anything other than mouse inputs as the shim is put into the orbiter mouse handler. 

I decided that I should just combine the PRs because they're both complements to each other and should be merged together

### DIFF
--- a/Src/Orbiter/Orbiter.cpp
+++ b/Src/Orbiter/Orbiter.cpp
@@ -2639,6 +2639,12 @@ void Orbiter::UserJoyInput_OnRunning (DIJOYSTATE2 *js)
 
 bool Orbiter::MouseEvent (UINT event, DWORD state, DWORD x, DWORD y)
 {
+	// Prioritizes mouse handling while in rotation mode
+	if (g_pOrbiter->StickyFocus()) {
+		if (event == WM_MOUSEMOVE) return false; // may be lifted later
+		if (g_camera->ProcessMouse(event, state, x, y, simkstate)) return true;
+	}
+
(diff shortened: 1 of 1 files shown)


jarmonik:
g_iCursorShowCount not initialized.

ZigTag:
> g_iCursorShowCount not initialized.

oh you're right, thats a silly mistake, let me fix that

ZigTag:
I ran this patch for around 6-7 hours, I think 5 hours straight, and it seems to be very stable. I found no issues throughout all of my gameplay. (I know I am the author, but I hadn't tested it this thoroughly before)

n7275 (approved):

PR #494, opened 2024-09-15
[Lua]Add some oapiGetPlanetxxx functions

TheGondos:
This PR add some planet related oapi functions:
-get_planetobliquity
-get_planettheta
-get_planetobliquitymatrix
-get_planetcurrentrotation
-planet_hasatmosphere
-get_planetatmparams
-get_groundvector
-get_windvector
-get_planetjcoeffcount
-get_planetjcoeff,

### DIFF
--- a/Script/Tests/GeneralApiTest.lua
+++ b/Script/Tests/GeneralApiTest.lua
@@ -20,6 +20,17 @@ function pass()
 end
 
 
+-- float comparison
+function almost_equal(a,b)
+	if a==0 and b==0 then
+		return true
+	end
+	if b==0 then
+		return math.abs(a) < 0.000000001
+	end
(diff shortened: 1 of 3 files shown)


DarkWanderer:
Any chance some tests could be added for these?

TheGondos:
I should be able to add some unit tests, except maybe for get_windvector as it probably depends on how the physics is configured

TheGondos:
I added some unit tests but I could not cover everything :
-get_planettheta, get_planetobliquitymatrix and get_planetcurrentrotation change over time, I have no clue how to validate this
-get_windvector requires activating the wind option, it's not something we can do from CTests
-get_planetjcoeffcount returns 0, even after activating non spherical gravity sources, I must be missing something...

PR #496, opened 2024-09-20
Splashscreen customization

TheGondos:
~~This PR should probably wait for after the 2024 release.~~
This adds support for custom load screens via a new scenario parameter "SplashScreen" in the ENVIRONMENT section :
```
BEGIN_ENVIRONMENT
   System Sol
   Date MJD 51982.0975169074
   SplashScreen lime Images/myawesomelogo.jpg
END_ENVIRONMENT
```

The first parameter is a color ~~in 0xbbggrr format~~ described as a [CSS color](https://drafts.csswg.org/css-color/#named-colors) (either in #hex or as a color name), followed by the image filename to be shown on the splashscreen.
The color name is case insensitive, I tested with filenames containing spaces and it works OK.
The image is scaled with the same logic as the embedded Orbiter logo.

Since I was messing with State.h, I took the opportunity to replace the error prone char arrays with std::string

End result :
splashscreen


maxq519:
Why should this wait for after the 2024 release?  Seems like a relatively minor feature in terms of code, but would be really cool to have.

TheGondos:
We need to freeze features or we'll never release a new version...
I'm not really satisfied with the syntax for the color but I don't want to investe more time in this before I get some feedback on the feature as a whole.

TheGondos:
With the 2024 version now out, it's time to get the ball rolling again,
I didn't find any document describing the .scn file format to add this to. Any idea where I could put it?

Xyon:
> With the 2024 version now out, it's time to get the ball rolling again, I didn't find any document describing the .scn file format to add this to. Any idea where I could put it?

https://github.com/orbitersim/orbiter/blob/main/Doc/Orbiter%20Developer%20Manual/SCENARIO.tex ?

TheGondos:
Rebased to latest main + documentation updated

Issue #512, opened 2024-11-21
Default "HOLD ALT" autopilot cannot be disabled


MishaPhysics:
I am noticing the following issue with Orbiter 2024 RC 2 (and with Orbiter 2024 RC 1, OpenOrbiter):

Lauch the default scenario /Delta-glider/Brighton Beach. Take off to some altitude with hover engines. Disable hover engines to start falling with some negative vertical speed. Enable HOLD ALT autopilot for two seconds (hover engines will enable) and disable the autopilot. Now hover engines are burning and you cannot disable it (The dot key (.) on numpad is ignored like the autopilot is enabled).

I do not notice this problem in Orbiter 2016.

TheGondos:
I think that there is something wrong in the thrusters code : there is a "level_permanent" field that is not saturated ([e.g. here](https://github.com/orbitersim/orbiter/blob/98870cb668cd03e3e5ce0dd8914b438dbaba5055/Src/Orbiter/Vessel.cpp#L8299)).
Adding some clamping between 0.0 and 1.0 seems to solve the issue but this part of the code is quite complicated so it would be nice if someone can confirm there are no side effects when doing that.

MishaPhysics:
Since this issue does not occur with Orbiter 2016 then could it mean that this issue appeared with the initial OpenOrbiter code? If I understand correctly it is not possible to look at Orbiter 2016 code...

MishaPhysics:
Thanks. I just tried the last build and I confirm that the issue disappeared.

PR #511, opened 2024-11-17
[Lua]Fix clbk_visualdestroyed not being called

TheGondos:
Fix lame typo in ScriptVessel.cpp preventing clbk_visualdestroyed from being called.

PR #513, opened 2024-11-21
Fix unchecked thruster level

TheGondos:
This PR fixes the issue #512
The "permanent" level of thrust was not clamped when using the IncThrusterLevel API.


TheGondos:
There was definitely a change in the CI environment causing build breakage. Had to cherry-picked the include fix from #511

PR #516, opened 2024-11-23
Galileo vessel fix

GLS-SSV:
Replaced non-existing Galileo vessel with Delta-Glider in the Jupiter flyby scenarios.


n7275 (approved):
This was broken as far back as 2006

GLS-SSV:
> This was broken as far back as 2006

I actually checked the releases (and meanwhile forgot the exact date), and these scenarios were added in 2004 or 2005

PR #535, opened 2025-01-01
2024 debuglibs

Xyon:


Xyon:
> @Xyon is this ready to be tested/merged?

No, not yet. The PR was just to get it to run the actions for testing - the x64 debug build is not successful from this branch atm.

GLS-SSV:
So this produces a release package and a debug package? Can't only one package be created with both lib types? Orbitersdk.lib and OrbitersdkD.lib

Xyon:
> So this produces a release package and a debug package? Can't only one package be created with both lib types? Orbitersdk.lib and OrbitersdkD.lib

My CMake skills have not yet got to the point where I know how to make it do that, but I would generally prefer that approach than a whole separated release, yes. For the O2024-compatible version, though, we can repackage the release but it might be well to offer a smaller "debug libs only" package for those who don't want to download the whole ~2.4GB package again just to get the debug libs.

GLS-SSV:
This solution of separate packages, from the point of view of the developer, is a PITA: either we'll need 2 separate environments with the same addon code (not practical), or we'll to have a new folder for the libs and manually copy the them to there, or rename the libs to keep all of them together (updating is error-prone).


This might help to put all in one package with different names: https://cmake.org/cmake/help/latest/guide/tutorial/Packaging%20Debug%20and%20Release.html

PR #550, opened 2025-01-15
WIP: Add support for ImGui

TheGondos:
This PR adds support for [Dear ImGui](https://github.com/ocornut/imgui), which provides an OS agnostic GUI toolkit.

- The first commit hooks the library to the orbiter core and the D3D9Client library. I've used cmake fetch+patch mecanism to add the external dependency. I'm not an expert in cmake so don't hesitate to comment if you're not satisfied with the way the library has been added.
The ImGui code is instanciated inside the Orbitersdk.lib file so that modules can eventually use it in the future.

- The second commit changes the handling of input boxes/context menus to use the new library.
Some feedback would be nice from users with high DPI monitors so that we can know if the font size is acceptable.

The library has backends for DX9/DX12, OpenGL, vulkan, WinAPI, SDL, glfw... so it should be future proof.

I've got ImGui equivalents for almost all dialogs in my linux branch so if we decide to expand on this, it should not be too difficult to replace the core Orbiter dialogs currently using the Windows API.

### DIFF
--- a/Extern/imgui/CMakeLists.txt
+++ b/Extern/imgui/CMakeLists.txt
@@ -1,0 +1,16 @@
+project(imgui)
+
+Include(FetchContent)
+
+FetchContent_Declare(
+  imgui
+  GIT_REPOSITORY https://github.com/ocornut/imgui.git
+  GIT_TAG v1.91.6-docking
+  PATCH_COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/imconfig.h .
+  UPDATE_DISCONNECTED 1
+)
+FetchContent_MakeAvailable(imgui)
+
+install(FILES ${imgui_SOURCE_DIR}/misc/fonts/Roboto-Medium.ttf
+	DESTINATION ${ORBITER_INSTALL_ROOT_DIR}
+)
--- a/Extern/CMakeLists.txt
+++ b/Extern/CMakeLists.txt
@@ -6,6 +6,7 @@ endif(NOT DEFINED ORBITER_BINARY_ROOT_DIR)
 
 add_subdirectory(lua)
 add_subdirectory(zlib)
+add_subdirectory(imgui)
 
 ## LFS
 add_library(lfs SHARED luafilesystem/src/lfs.c)
--- a/Extern/imgui/imconfig.h
+++ b/Extern/imgui/imconfig.h
@@ -1,0 +1,155 @@
+//-----------------------------------------------------------------------------
+// DEAR IMGUI COMPILE-TIME OPTIONS
+// Runtime options (clipboard callbacks, enabling various features, etc.) can generally be set via the ImGuiIO structure.
+// You can use ImGui::SetAllocatorFunctions() before calling ImGui::CreateContext() to rewire memory allocation functions.
+//-----------------------------------------------------------------------------
+// A) You may edit imconfig.h (and not overwrite it when updating Dear ImGui, or maintain a patch/rebased branch with your modifications to it)
+// B) or '#define IMGUI_USER_CONFIG "my_imgui_config.h"' in your project and then add directives in your own file without touching this template.
+//-----------------------------------------------------------------------------
+// You need to make sure that configuration settings are defined consistently _everywhere_ Dear ImGui is used, which include the imgui*.cpp
+// files but also _any_ of your code that uses Dear ImGui. This is because some compile-time options have an affect on data structures.
+// Defining those options in imconfig.h will ensure every compilation unit gets to see the same data structure layouts.
+// Call IMGUI_CHECKVERSION() from your .cpp file to verify that the data structures your files are using are matching the ones imgui.cpp is using.
+//-----------------------------------------------------------------------------
+
+#pragma once
+
+//---- Define assertion handler. Defaults to calling assert().
+// If your macro uses multiple statements, make sure is enclosed in a 'do { .. } while (0)' block so it can be used as a single statement.
+//#define IM_ASSERT(_EXPR)  MyAssert(_EXPR)
+//#define IM_ASSERT(_EXPR)  ((void)(_EXPR))     // Disable asserts
+
+//---- Define attributes of all API symbols declarations, e.g. for DLL under Windows
+// Using Dear ImGui via a shared library is not recommended, because of function call overhead and because we don't guarantee backward nor forward ABI compatibility.
+// - Windows DLL users: heaps and globals are not shared across DLL boundaries! You will need to call SetCurrentContext() + SetAllocatorFunctions()
+//   for each static/DLL boundary you are calling from. Read "Context and Memory Allocators" section of imgui.cpp for more details.
+//#define IMGUI_API __declspec(dllexport)                   // MSVC Windows: DLL export
+//#define IMGUI_API __declspec(dllimport)                   // MSVC Windows: DLL import
+//#define IMGUI_API __attribute__((visibility("default")))  // GCC/Clang: override visibility when set is hidden
+
+//---- Don't define obsolete functions/enums/behaviors. Consider enabling from time to time after updating to clean your code of obsolete function/names.
+//#define IMGUI_DISABLE_OBSOLETE_FUNCTIONS
+
+//---- Disable all of Dear ImGui or don't implement standard windows/tools.
+// It is very strongly recommended to NOT disable the demo windows and debug tool during development. They are extremely useful in day to day work. Please read comments in imgui_demo.cpp.
+//#define IMGUI_DISABLE                                     // Disable everything: all headers and source files will be empty.
+//#define IMGUI_DISABLE_DEMO_WINDOWS                        // Disable demo windows: ShowDemoWindow()/ShowStyleEditor() will be empty.
+//#define IMGUI_DISABLE_DEBUG_TOOLS                         // Disable metrics/debugger and other debug tools: ShowMetricsWindow(), ShowDebugLogWindow() and ShowIDStackToolWindow() will be empty.
+
+//---- Don't implement some functions to reduce linkage requirements.
+//#define IMGUI_DISABLE_WIN32_DEFAULT_CLIPBOARD_FUNCTIONS   // [Win32] Don't implement default clipboard handler. Won't use and link with OpenClipboard/GetClipboardData/CloseClipboard etc. (user32.lib/.a, kernel32.lib/.a)
+//#define IMGUI_ENABLE_WIN32_DEFAULT_IME_FUNCTIONS          // [Win32] [Default with Visual Studio] Implement default IME handler (require imm32.lib/.a, auto-link for Visual Studio, -limm32 on command-line for MinGW)
+//#define IMGUI_DISABLE_WIN32_DEFAULT_IME_FUNCTIONS         // [Win32] [Default with non-Visual Studio compilers] Don't implement default IME handler (won't require imm32.lib/.a)
+//#define IMGUI_DISABLE_WIN32_FUNCTIONS                     // [Win32] Won't use and link with any Win32 function (clipboard, IME).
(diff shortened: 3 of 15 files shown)


TheGondos:
I've added a temporary port for the D3D9 gfx dialog to help in refining the API. I'll either remove it later or port the other dialogs.
So far I'm not satisfied with the name DialogImGui, I'll probably switch to ImGuiDialog.
I also consider moving the window creation (ImGui::Begin) inside the base class to have a cleaner handling of the active flag which is used to track if the window is opened or closed.

jarmonik:
Thanks, I tested this and it worked fine, no problems with building and the code looks very simple and efficient. So, I'll have to start looking into the documentation and learning how to use the widgets. There seems to be pretty good collection of them. Is there a "Load file", "Save file" dialogs ? Is there still much to do before merge ?

TheGondos:
> Thanks, I tested this and it worked fine, no problems with building and the code looks very simple and efficient. So, I'll have to start looking into the documentation and learning how to use the widgets. There seems to be pretty good collection of them. Is there a "Load file", "Save file" dialogs ? Is there still much to do before merge ?

There is no "native" file dialog but there are a few projects that provide one, I'll try a couple and so if one is easier to use than the others.
You can place an "ImGui::ShowDemoWindow()" in the Draw() method to see a complete list of available controls. Most are easy to use, it becomes more complicated when you want an advanced layout but it usually can be done. For debug menus it's a no brainer to use.

Before merging I'd like some feedback from someone with a HiDPI display to know if the dialogs aren't too small there.
For now only the context menu / input box is using this. Did you try it (by using the TGT button in OrbitMFD for example)?

jarmonik:
Yes, I tried the "TGT" button and it works perfectly. Nice work there.  On 1920x1080 the scaling looks good al-tough in some cases I might prefer little smaller, like in a property tree. Is the scaling adjustable ?

TheGondos:
The font used to be 14 in size, I upped it to 18 but that's something that'll need tweaking I guess.
You can modify this line in DlgMgr.cpp if you want to play with different values :
`	io.Fonts->AddFontFromFileTTF("Roboto-Medium.ttf", 18.0f, &config, GetGlyphRangesOrbiter());
`
I see there is a dlgFont_Scale configuration option, I'll see if I can make use of it.

TheGondos:
This is ready for review **but not for merge**.
If you're OK with this, just tell me and I'll remove the D3D9 Graphics Controls prototype and squash the whole thing so it won't pollute the main branch.

TheGondos:
Well, since I was at it I added notifications and used them for Lua errors :
notif

TheGondos:
I've encountered some bugs when working on implementing the core dialogs with ImGui.
I'll backport the fixes here when I'm satisfied

Extern/imgui/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+  GIT_REPOSITORY https://github.com/ocornut/imgui.git
+  GIT_TAG v1.91.7-docking
+  PATCH_COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/imconfig.h ${CMAKE_CURRENT_SOURCE_DIR}/IconsFontAwesome6.h .
+  UPDATE_DISCONNECTED 1
+)
+FetchContent_MakeAvailable(imgui)
+
+install(FILES ${imgui_SOURCE_DIR}/misc/fonts/Roboto-Medium.ttf fa-solid-900.ttf
+	DESTINATION ${ORBITER_INSTALL_ROOT_DIR}
+)
```
ThePuzzlemaker, line 16: I'd recommend adding a `file(COPY ...)` here; running in CLion for me under a debugger doesn't work since the files aren't `install`ed with that. Not sure what VS does but I can't be bothered to test it :P

OVP/D3D9Client/DebugControls.cpp
```diff
@@ -269,7 +297,9 @@ void Release()
 {
 	vObj = NULL;
 	hDlg = NULL;
-	hGfxDlg = NULL;
+	oapiCloseDialog(gfxDlg);
```
ThePuzzlemaker, line 300: This is causing an access violation exception when closing the launchpad as `g_pOrbiter->DlgMgr()` appears to be a `nullptr` at this point as it is deleted in `Orbiter::CloseSession()`. This isn't a *particular* issue since it's on exit, but it is quite annoying when using a debugger, and still could lead to issues in the future.

TheGondos: yes I fixed it ;-) I pushed the other branch for the ImGui convertion, it's available here if you want to have a look : https://github.com/TheGondos/orbiter/tree/imgui_dlg

jarmonik:
I did some testing and couldn't find anything wrong. Is this ready for merge ?

TheGondos:
> I did some testing and couldn't find anything wrong. Is this ready for merge ?

Not quite yet, I hope to have this ready this weekend

jarmonik:
> Not quite yet, I hope to have this ready this weekend

Ok, take your time, there's no rush. There are some other things I can do meanwhile.

OVP/D3D9Client/D3D9Pad.cpp
```diff
@@ -1304,8 +1304,9 @@ void D3D9Pad::Polygon (const IVECTOR2 *pt, int npt)
 #endif
 
 	if (npt<3) return;
-	if (HasBrush() && npt > 64) return;
-
+	// The VectorMap drawing code creates polygons with at least 68 points
+	// Let's accept up to 70
+	if (HasBrush() && npt > 70) return;
```
TheGondos, line 1309: @jarmonik : I had to increase the max value for some polygons, are there constants to change somewhere else because of this?

TheGondos:
Added ImGui version of External MFD too because the context menu would open underneath the WinAPI window.

OVP/D3D9Client/D3D9Pad.cpp
jarmonik: I don't recall other places, so, it's likely just that. It's ok to increase it. But the polygon triangulation is N3 process so it can get performance heavy quickly. It's recommended for higher poly counts to create a poly object and use SketchPad::DrawPoly() especially if the polygon shape is a constant. (i.e. no per frame triangulation)

TheGondos: OK thanks, does it handle filled concave polys?

Extern/imgui/CMakeLists.txt
TheGondos: This should be addressed globally if there is an issue with the install primitive

OVP/D3D9Client/D3D9Pad.cpp
jarmonik: Yes it does fill concave polys.

jarmonik:
What's the status of this ?  Ready for merge ? Is there something that's still missing ?

TheGondos:
> What's the status of this ? Ready for merge ? Is there something that's still missing ?

I squashed everything, it should be ready to merge. I'd feel better if someone can test this before merging though...

n7275:
> > I can try building and testing later today. Anything in particular I should look out for?
> 
> Thanks! The main visible side effects are a context menu/input box instead of the old Select/InputBox dialogs when e.g. selecting a target in an MFD. If you have a multi-monitor setup, I'd be interested to know if there are issues when the context menu expands over the other monitor. Also if you have a HiDPI screen, you can check if the size of the controls are OK. The external MFD has been ported so you can also try to move it across monitors to see if it behaves correctly. I have no idea how it'll behave if you drag it on monitors with different DPI.

I have a single monitor, but it's 21:9, which is its own set of edge-cases.

Two questions:
* Is this supposed to be implemented for all menus, or just a few (D3D9 controls, the MDF input menus).
*When I click TGT in Orbit MFD (for example) the dialogue menu appears where I clicked, when I try the same thing in external MFD, the box appears at the lower-left corner of the MFD, not near my cursor. Is this the intended behavior?

I tested a lot in windowed mode too, and was not able to make the menu go off screen, which is really nice. They extend outside of the Orbiter Window, but not off screen.

Other than that, I wasn't able to break anything.

TheGondos:
>  Is this supposed to be implemented for all menus, or just a few (D3D9 controls, the MDF input menus).

There is [another branch](https://github.com/TheGondos/orbiter/tree/imgui_dlg) to port other dialogs that I'll rebase once this one is merged.

>   When I click TGT in Orbit MFD (for example) the dialogue menu appears where I clicked, when I try the same thing in external MFD, the box appears at the lower-left corner of the MFD, not near my cursor. Is this the intended behavior?

No, that should not happen. It should spawn where you click :
image

n7275:
> > Is this supposed to be implemented for all menus, or just a few (D3D9 controls, the MDF input menus).
> 
> There is [another branch](https://github.com/TheGondos/orbiter/tree/imgui_dlg) to port other dialogs that I'll rebase once this one is merged.
> 
> > When I click TGT in Orbit MFD (for example) the dialogue menu appears where I clicked, when I try the same thing in external MFD, the box appears at the lower-left corner of the MFD, not near my cursor. Is this the intended behavior?
> 
> No, that should not happen. It should spawn where you click : image
> 
> Edit : @n7275 : I tried to reproduce your issue in windowed or fullscreen mode but I'm failing miserably. What exact steps are you using?

Sorry this took a bit to get back to. I think I must've had the D3D9 external MFD open before. I have no issue with the regular MFD...but that bring up another point.

I really like that I can make ExternalMFD any aspect ratio, but it would be really nice if we could lock it to a specific ratio, or set it interactively, specifically because we use it a lot for NASSP with MOCR displays. The minimize feature is amazing BTW, love it. Other than that small feature request on my part I see no reason this can't be merged.

jarmonik:
> > I think I must've had the D3D9 external MFD open before.
> 
> That makes sense, that's the reason I converted the external MFD in this branch to ImGui so that the context menu shows OK. Makes me wonder how many addons could be having the same issue...

That is strange. D3D9 ExtMFD opens a secondary swap-chain for drawing, so, maybe ImGUI can't draw on a top of it.  But since the ImGUI is running on D3D9 it might offer better alternatives for a secondary swap-chain, maybe there could be ImGUI control/element/widget for direct Sketchpad drawing making Sketchpading easier for users.

I am currently working on Vulkan renderer and it's very likely that D3D9 will be discontinued after it's in a working condition. So, no point in putting time and efforts in D3D9 specific problems. So, solving the D3D9 ExtMFD issue can wait for Vulkan.

TheGondos:
> > > I think I must've had the D3D9 external MFD open before.
> > 
> > 
> > That makes sense, that's the reason I converted the external MFD in this branch to ImGui so that the context menu shows OK. Makes me wonder how many addons could be having the same issue...
> 
> That is strange. D3D9 ExtMFD opens a secondary swap-chain for drawing, so, maybe ImGUI can't draw on a top of it. But since the ImGUI is running on D3D9 it might offer better alternatives for a secondary swap-chain, maybe there could be ImGUI control/element/widget for direct Sketchpad drawing making Sketchpading easier for users.
> 
> I am currently working on Vulkan renderer and it's very likely that D3D9 will be discontinued after it's in a working condition. So, no point in putting time and efforts in D3D9 specific problems. So, solving the D3D9 ExtMFD issue can wait for Vulkan.

Actually that's to be expected : when ImGui "sees" that the popup menu can be contained in the main window, it does not create a new native window, it just renders the dialog with a draw list on the main "viewport". Looks like there's no workaround for this (cf [this issue](https://github.com/ocornut/imgui/issues/7105)).

Xyon:
What's the status of this? Ready to merge, or more testing needed? I'd like to drive down the number of open PRs before the year ends.

TheGondos:
> What's the status of this? Ready to merge, or more testing needed? I'd like to drive down the number of open PRs before the year ends.

It's OK with me, just waiting for @n7275 to confirm the last modifications I made to take his remarks into account are adequate.

Xyon:
@n7275 are you happy with merging this as-is?

Xyon:
@TheGondos can you rebase the conflicts, we can get this merged in then.

TheGondos:
@Xyon : done

PR #553, opened 2025-01-22
Linux CI and basic dependency management

DarkWanderer:
This PR contains GitHub workflows and CMake changes needed to start an "official" Linux branch in Orbiter repo. The codebase itself does not compile yet - there is a number of errors still to fix, most notably the unmaintainable CMakeLists mess around LDoc/LFS. However, it is enough to allow cooperative work on Linux in central location

There are 2 options to incorporate this PR:

## Option 1: Merge to `main`

If this change is merged to `main`, GitHub actions will start building Linux branch for each PR. It will be inconvenient as it will flag failures for good PRs, but it will allow to iterate the quickest

## Option 2: Merge to separate `linux` branch

This way, work on finalizing Linux build can be kept separate from `main`, at the cost of potential desync with `main` branch and associated manual labor required to keep `linux` branch up to date with `main`

### DIFF
--- a/.github/workflows/on-push-linux.yml
+++ b/.github/workflows/on-push-linux.yml
@@ -3,18 +3,29 @@ name: On Push (Linux)
 on:  
   push:
     branches:
+      - main
       - linux
     paths-ignore: [ 'Doc/**', 'Flights/**', 'Html/**', 'Images/**', 'Localdoc/**', 'Scenarios/**', 'Textures/**', 'Textures2/**']
   pull_request:
     branches:
+      - main
       - linux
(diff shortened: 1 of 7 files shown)


n7275:
I don't have a super strong opinion on this, but I think as long as someone "owns" the linux branch (in a project management sense) I think option 2 is fine, with the goal of getting to a minimal build able output, even if it doesn't yet have a graphics client.

jarmonik:
Last I had linux on my computer was about 20 years ago. It's been by intention to try FreeBSD but haven't got enough inspiration to risk wrecking a working computer. So, I don't know much about linux development. If I am writing a code that I know is not linux compatible then it could be isolated via "#ifndef __LINUX" but then again DX9 stuff shouldn't be linux compatible and there's lot of it. If linux stuff is merged to "main" I can't hold off commits and PR just because build it fails on linux due to reasons I don't know why. So, it would be up to a linux specialist to resolve the issues later on.

Issue #558, opened 2025-02-09
D3D9Pad::TextW does not use vertical and horizontal text alignment from SetTextAlign


indy91:
I have been playing around with TextW, to be able to print greek characters in MFDs, and I noticed that TextW does not seem to honor the horizontal text alignment set with SetTextAlign. Looking at the code in D3D9Pad.cpp and D3D9Pad2.cpp, TextW does not seem to use the variables tah and tav that would be used for the text alignment.

TheGondos:
I can take a look. If it was only me I'd ditch TextW and add UTF-8 support to Text. Maybe we'll do that later

indy91:
Thanks for taking a look! I definitely don't insist on keeping TextW, but I would like there to be a capability to easily print greek letters and subscript/superscript on MFDs, that is what I was using TextW for.

TheGondos:
I pushed something for this. Please note that text rotation is not supported.

jarmonik:
Fixed.  Should be possible to rotate with SetWorldTransform()

PR #559, opened 2025-02-12
[D3D][Fix]Honor text alignment in TextW

TheGondos:
This is a tentative fix for #558
@jarmonik : can you have a look at this ? Looks like the color format is wrong for TextW. I don't know if it's configurable somewhere so I had to invert the red and blue channels.


jarmonik (approved):
It looks good to me. Although, It's a bit strange that incorrect color has not been noticed earlier.
SkpColor class does some swaps between R-B, so it's a known problem.  Maybe DrawTextW() does it's own swap. Let's fix the problem as you have suggested, there's no point going through the entire Sketchpad interface at this point. I'll do a cleanup and overhaul of the sketchpad interface when switching to Vulkan. 

Go for merge.  ( Is it ready ? )

TheGondos:
Yes it's ready

PR #561, opened 2025-02-19
SDL3 port for cross-platform support

ThePuzzlemaker:
# Preface

I am absolutely willing to split this PR into many as necessary to ease review once I am done with it, and I am of the understanding that this PR will probably not land as-is once complete. I am not a C++ programmer by trade or by hobby (I mostly do Rust code) so if there's anything you see that's a bit, well, _icky_, just let me know and I am willing to fix it.

This PR builds on #550, and on @TheGondos's `imgui_dlg` branch which extends on #550 further. If/when #550 and associated dialog changes are merged, I will rebase this PR onto that branch.

# Rationale

Orbiter does not work well on Linux as of the moment, only partially working with WINE and dxvk. It requires a lot of changes to undo 25+ years of Windows-only development, but this can be changed. Using SDL3 (which is the recommended version of SDL by its developers), many of the hard things of cross-platform support, including module loading, window initialization, input support, etc., become easy API calls to a relatively lightweight translation layer. SDL3, and in extension SDL as a whole, have been battle-tested in the game development industry (for example, SDL2 is used in many Valve titles).

# Methodology

This PR is (and will be) huge.

Breaking changes are made in some cases, and I believe this is justified since the x64 release will break binary compatibility anyway, so we might as well merge some additional minor breaking changes at the same time to reduce stress on module/vessel maintainers. [**Update 2025-02-22**: I've realized I've probably made too many breaking changes right now, so I think many of these will go into a different "Cross-platform" PR.]

D3D9Client will **not** be deprecated. I plan to work on my own Vulkan client after I'm done with this, but I will probably never reach the level of @jarmonik's amazing work on D3D9Client. Instead, D3D9Client will be changed as part of this or another PR to use cross-platform APIs for initialization and interaction with the Orbiter core, but ultimately will still use D3D9 and Windows-specific APIs under the hood, and just won't be able to be built on non-Windows targets.

# Breaking Changes

## Build

`CMAKE_CXX_STANDARD_REQUIRED` is now `ON` -- C++17 conformance is required by the building compiler. MSVC has supported almost all features of C++17 (except one, `std::hash<std::filesystem::path>`, which is not currently used in this codebase) since VS2017 version 15.7 (compiler version 19.14.26428.0). Thus, you don't need to have a very up-to-date compiler to still build Orbiter.

Modules must now link with SDL3. If building with Orbiter's CMake system, this is as simple as adding `SDL3::SDL3` to the `target_link_libraries` invocation. I tried making it so this was not necessary but it was non-trivial and would've led to other issues down the line.

## SDKs

- `GraphicsAPI.h`: Render windows are now SDL wrappers. I have yet to add "proper" docs but these classes are fairly self-explanatory.
- `ModuleAPI.h`: The signature of `clbkProcessMouse` has changed to take SDL mouse events rather than WinAPI event types. The conversion is fairly straightforward.

# Appendix

I want to thank all the Orbiter developers for their excellent work making this simulator. I hope that this PR can help continue to make Orbiter better and even more accessible to people. I understand that this PR, once finished, will likely be a large burden on reviewers, and I will do whatever I can to ease such burden. And again, I'm not particularly steadfast on any code here. If you ask me to change something, I will almost certainly be willing to change it.

Up to date as of 2025-02-26, changes up to 3e1c67c

### DIFF
--- a/OVP/D3D9Client/WindowMgr.cpp
+++ b/OVP/D3D9Client/WindowMgr.cpp
@@ -1071,12 +1071,34 @@ bool WindowManager::MainWindowProc(const SDL_Event &event)
     return false;
 }
 
+bool WindowManager::MainWindowProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
+{
+	static int xpos, ypos;
+	switch (uMsg) {
 
+	case WM_MOUSELEAVE:
+	case WM_MBUTTONDOWN:
+	case WM_LBUTTONDOWN:
+	case WM_LBUTTONUP:
+		return false;
 
+	case WM_KEYDOWN:
+	{
+		return false;
+	}
 
+	case WM_MOUSEWHEEL:
+		return false;
 
+	case WM_MOUSEMOVE:
+		xpos = GET_X_LPARAM(lParam);
+		ypos = GET_Y_LPARAM(lParam);
+		MouseMoved(xpos, ypos);
+		break;
+	}
 
-
+	return false;
+}
 
 // ===============================================================================================
 // SideBar Implementation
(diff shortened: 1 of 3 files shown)


jarmonik:
Sounds good to me. However, I would not merge MFD and MFD2 classes because there are plenty of MFD's using the old windows GDI interface. As far as I can tell, there are no problems continuing to support it on windows. Linux could simply deal with the problem by not loading modules those are using GDI.  

Also the GDI is extensively used by large projects like NASSP in other aspects beyond MFD's, so, might be good idea the talk to them before discontinuing GDI support, I try to listen developers and their wishes,

If the GDI support is continued on windows what harm does it do on linux ?

ThePuzzlemaker:
> If the GDI support is continued on windows what harm does it do on linux ?

Ultimately, I guess it's not that much harm -- it's just a lot of `#ifdef`s would be required which would clutter things up, but I understand your reasoning.

I would hope that the Sketchpad system eventually becomes fully fleshed out so we can deprecate GDI usage and make everything fully cross-platform. If you have any questions about APIs, I'd suggest looking at [DearImGui's DrawList](https://github.com/ocornut/imgui/blob/773060155fe7dd2c8d71751f99543fbf045d6bf4/imgui.h#L3057-L3366) which doesn't seem to have too bad an API; it might be useful as a reference to flesh out the Sketchpad API so that it supports whatever devs need.

But yes, completely removing GDI without first _properly_ deprecating and superseding it is not a good idea, so I'll work on those changes (including reverting the MFD/MFD2 merge in 0109d67040bf6059ba8d3d6a303677f9a767c98b) once I'm finished getting D3D9Client working again (currently it's having some odd issues, and I'm not exactly sure where they're coming from, so I'll have to really scrutinize my code)

ThePuzzlemaker:
I think once I get everything working again I'll probably split the Launchpad into a different PR.

ThePuzzlemaker:
D3D9Client is not entirely working yet... see [this reply](https://www.orbiter-forum.com/threads/draft-pr-sdl3-port-for-cross-platform-support.42146/#post-627357) on the forum

jarmonik:
I browsed through the modifications I noticed a lot of changes that doesn't seem to be relevant to SDL such as scenarios, documentation, *.jpg's and *.ico's. Are those slipped there by accident ?  It's pretty hard to make sense out of it. Also, there are plenty of white space changes. I always check my code before commit and revert white space changes and I hope you could too. Changes in white space just make spotting relevant changes a little more difficult, increasing work load.

ThePuzzlemaker:
The scenario and documentation changes are relevant, but more as part of the Launchpad changes--that will be split into another PR.

As for the texture stuff, that's cause I moved the location of the Launchpad bitmaps to be in the Textures folder as this is easier than finding a (eventually cross-platform) way to store and load resources within a binary. (Not until `#embed` in C++23 at least, I guess)

And the whitespace changes are mostly incidental, cause my editor refuses to save files with trailing whitespace. I'll make sure any irrelevant whitespace changes get removed.

ThePuzzlemaker:
I've squashed the previous some-odd commits because it made it easier to revert unnecessary changes which will be split into a future separate launchpad PR.

ThePuzzlemaker:
It appears my editor has decided to continue removing trailing whitespace. I think this is because of the `.editorconfig` at the root of the repo. I'll get to removing the extra whitespace changes soon.

It might be worth making a PR using something like `clang-format` which will automatically format code to a specific (quite customizable) style. But that's somewhat irrelevant..

ThePuzzlemaker:
Mouse input is currently broken (at least for me) -- for some reason only every other right click is detected. I found a forum post about this on the SDL forums but it's from 2002... I highly doubt a bug lasted that long, so I'm pretty sure I'm doing something wrong. I've asked the people of the SDL Discord so we'll see I guess

As for keyboard input, it is working satisfactorily. I'm not sure if this method is the *best* way of doing things (using a linked list to buffer events since SDL doesn't buffer them like DirectInput does), but it's certainly the easiest and least invasive way.

ThePuzzlemaker:
This PR is now ready for review, **but will not be ready for merge until #550 (et seq.) are merged**

This will probably need some shakeout testing to ensure it doesn't silently break certain things, but major functionality which was touched by this code is appearing to work well for me.

There is one unresolved FIXME, in Src/Orbiter/Orbiter.cpp:2004, which I'm open for suggestions on how to resolve; ultimately it's hard to simply force-destroy a `std::shared_ptr<T>` without crashing. I think a saner option might be to run CloseSession but I'm not familiar enough with Orbiter core code to know if that would cause issues if KillVessels failed.

DragonSWDev:
As for the D3D9Client what about using DXVK that can be built as native Linux library and used for providing Direct3D API implemented on top of Vulkan on Linux? I think Valve uses it on some of their older Source 1 based games. Probably some changes will be needed but I guess it should be easier than writing new client from scratch.

DragonSWDev:
@ThePuzzlemaker Maybe things will be different if Direct3D 9 code will be compiled natively to Linux with DXVK native?

jarmonik:
Based on my understanding merging this pull request would result a brand new Orbiter where nothing previously existing would work. Only the stock add-ons. So, how the community would take that ?

When the time comes we could try to release two Orbiter versions one with the legacy interface and another one with the SDL giving developers more time to adjust.

ThePuzzlemaker:
@jarmonik No. This pull request has been made with minimal breaking changes in mind; at the moment, it is still very possible to deal with Win32 APIs directly instead of ImGui for windowing.

Originally that was not the case, but I realized I was being too ambitious and that changes need to be made in smaller increments. All that this PR is _intended_ to do (whether it does so at the moment I would need to look at the code closer again) is to reduce the usage of Win32 APIs in favor of the cross-platform SDL to lead Orbiter towards a potential cross-platform future. This mostly means windowing, which only really touches stuff that was changed to ImGui now, as well as input, which I'll definitely want to coordinate with any others (like @TheGondos, if I recall correctly) as I believe there are some working on improved input for Orbiter. 

It still needs to be updated with @TheGondos's ImGui changes (that have been merged) over the last year or so but I've been super busy with starting out university (in my second semester now) and haven't had the motivation to do so yet. This may end up removing some or all of the breaking changes as I plan to completely rebuild this PR when I get the time. So might not be until during the summer, to be honest (I have prior commitments during my spring break).

Anyone else is welcome to take a shot at it until then (though if you use my code directly I would like credit :P)

PR #580, opened 2025-03-18
Always create surface tile labels on initial load, regardless of launchpad config

ThatRedMelon:
Whenever a surface tile is loaded, it first checks if the surface markers setting is enabled before creating the labels for the tile, meaning that if the setting is enabled during a session, previously loaded tiles don't show their labels, as they haven't been created. This PR removes the check in order to always create the labels for a tile, regardless of the setting. This doesn't affect the actual display of the labels, which still honors the setting.


rcflyinghokie:
Can confirm this resolves the issue

PR #606, opened 2025-10-19
Use std::chrono instead of Windows PerformanceCounter

TheGondos:
This PR replaces the Windows only QueryPerformanceCounter/QueryPerformanceFrequency API with C++ chrono equivalents.

### DIFF
--- a/Src/Orbiter/Log.cpp
+++ b/Src/Orbiter/Log.cpp
@@ -415,33 +415,25 @@ void tracenew (char *fname, int line)
 // =======================================================================
 
 static double prof_sum = 0.0;
-static double prof_scale = 0.0;
-static DWORD prof_count = 0;
-static LARGE_INTEGER prof_t0;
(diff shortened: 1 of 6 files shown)


TheGondos:
This has been tested indirectly in my linux branch since I basically copy pasted from it. I redid some basic tests on this branch and could not see any change in behavior.
A bunch of code was removed that was checking if precise timing was available or not. I assume that's not a problem anymore with "recent" PCs.

TheGondos:
Just to be clear, the change is related to the system time, not the simulated time. A bug would involve some kind of time dilation/contraction between clock time and simulated time, it would not show when comparing position/velocity against simulated time.
The one point I have some doubt is the behavior when the computer is going in suspend mode and waking up.
Looking at the SDL API, I saw no way to be notified of these events so I'm not sure it's something we can support in a portable way anyway.

PR #610, opened 2025-11-08
[POC]Virtual filesystem

TheGondos:
This is a prototype/proof of concept for a virtual filesystem.
The idea is to be able to deploy addons alongside Orbiter instead of having to extract files directly in the installation directory.
It is modeled to behave loosely like an overlayfs in linux : you declare a write directory where any file update will be made, and you can add overlays that will be scanned when you try to open files.
I reflected on using PhysicsFS but loading DLLs from archive files is not possible, so this makes 90% of it useless.
There is still a lot to do (Lua compatibility, file suppression, bug hunting/fixing, configuration) but if you have any idea that can help, feel free to comment here.


TheGondos:
The last update will look for directories in an optional "Addons" directory inside the Orbiter installation and add an overlay for each one it founds.
I tested with planets .tree file, DLL addons and Lua modules.
Some work but some that read/write files directly don't play well with this (e.g. IMFD fails because it searches for a config file)

PR #614, opened 2025-12-08
Port Orbiter dialogs to ImGui

TheGondos:
This PR modifies the dialogs to use ImGui instead of the Win32 API.
The "MenuInfoBar" (menu on the top) has been changed to allow addons to add their own button, it's now fully dynamic and buttons can be reordered/removed with drag and drop.
Some tests on a HiDPI screen are needed to see if some rescaling is needed (I don't have one handy).

### DIFF
--- a/Src/Orbiter/DlgMenuCfg.cpp
+++ b/Src/Orbiter/DlgMenuCfg.cpp
@@ -30,6 +30,7 @@ void DlgMenuCfg::OnDraw()
 
 	if(ImGui::BeginTabBar("##MenuCfgTab")) {
 		if(ImGui::BeginTabItem("Appearance")) {
+			ImGui::BeginChild("##Child");
 			ImGui::SeparatorText("Menu bar");
 			ImGui::PushID(1);
 			ImGui::RadioButton("Show", &prm.MenuMode, 0);
@@ -61,6 +62,7 @@ void DlgMenuCfg::OnDraw()
 			const char *fpsmode[]={"None", "On the left", "On the right"};
 			ImGui::Combo("##FPSMode", &prm.FPS, fpsmode, IM_ARRAYSIZE(fpsmode));
 
+			ImGui::EndChild();
 			ImGui::EndTabItem();
 		}
 		if(ImGui::BeginTabItem("Menu items")) {


TheGondos:
> Yes, it's fine by me. I don't know how the TextW works internally since it relies on build-in DX9 feature. It should work so UTF-8 to WCHAR conversion should be ok.

@jarmonik : Ok good to know, I think it's wiser to do that in another PR, there's already enough ways this PR can break things.
@Xyon : I added a manual workflow to help with providing test binaries. I had to change the condition for the pack target and changed the retention to 14 days. Is it OK with you?

Xyon:
Works for me. Do you want to cherry pick that into its own feature so we can merge that in independently of the UI overhaul?

Xyon:
Alright. This is quite a large overhaul and a lot of changes, so while I'm quite keen on these changes coming in I think I'm going to have to try to do some testing on my end as well to make sure the end result is stable, especially with other addons in the mix that might bring their own dialog stuff. I imagine there won't be much conflict but I'd rather not just assume it'll be fine.

I don't have a good idea of what that'll look like right now, though, other than building the branch and manually poking about with it. :thinking:

TheGondos:
> I don't have a good idea of what that'll look like right now, though, other than building the branch and manually poking about with it. 🤔

There are prebuilt binaries here : https://github.com/TheGondos/orbiter/actions/runs/20211823089

Xyon:
Okay. I'm content it builds and I've not been able to make it break anything. I asked some of the NASSP guys to have a go of it, to sense check what kind of impact it might have on a large complex addon, and they're reporting similarly few issues. On the whole I think we can merge this and fix-forward any lingering issues with subsequent PRs. One thing that did stick out as a slight omission was the Scenario Editor having the older UI still, which is probably because its in its own module and therefore not really a "core" dialogue per-se. But I'm happy to bring this in if you're happy it's ready as-is, @TheGondos

Src/Orbiter/Config.h
```diff
@@ -219,24 +219,25 @@ struct CFG_DEVPRM {
 
 struct CFG_JOYSTICKPRM {
 	DWORD  Joy_idx;				// joystick device index (0=disabled)
-	DWORD  Deadzone;			// central deadzone range for all axes (0-10000)
+	int    Deadzone;			// central deadzone range for all axes (0-10000)
 	DWORD  ThrottleAxis;		// joystick throttle axis (0=none, 1=z-axis, 2=slider 0, 3=slider 1)
```
Xyon: Can these ones move to `int` as well, if we're removing the DWORDS for the others? Or is there a reason to keep them this way?

PR #627, opened 2026-01-03
Use miniaudio instead of irrKlang

TheGondos:
This is another shot at a replacement for irrKlang since SDL3_mixer looks problematic.
This one uses miniaudio.
There are still references to SDL3_mixer since I started from that branch. Will clean this up if we have a winner.


dbeachy1:
@TheGondos I'm curious what the larger reason is to swap IrrKlang Pro for another library -- this looks like a major change, and it will need extensive testing in any case before we would merge it.

PR #643, opened 2026-04-27
Remove dead GDI related code

TheGondos:
I just tried removing the GDIClient.cpp/GDIClient.h files and it turns out they are no longer required to build Orbiter.
This PR cleans them up, along with a few obsolete members in the Orbiter core (Pane).
The change to GraphicsAPI.h is purely cosmetic.


n7275:
@Xyon this can be merged too

PR #646, opened 2026-05-02
(XRSound) Tri-state RCS sounds and updated assets

EvESpirit:
Hello again!

@fconcas (fcon_91 on Discord) came up with the idea of a tri-state RCS change for the XRSound package, and produced some gorgeous raw assets for the respective RCS Attack (Open), Sustain and Close slots.

This patch series implements a new slot for RCS sound playback in a backwards-compatible way (no mods will break if a vessel isn't configured for this change, and the old RCS Attack (Open) + RCS Close system is used as it was before it - albeit with the new sound assets, that change is overriding).

_Preview:_

[Preview](https://www.youtube.com/watch?v=QCP5TaIx2jk)


n7275:
I have not had a chance to review the code fully yet, but from what I've looked through and the demo, this is awesome! Most of what you actually hear with RCS jets inside the spacecraft is the solenoid valve actuating when it opens or closes. I've wanted to add the on-sustain-off feature to NASSP for a very long time. Thank you.

I will review the code fully very soon.

EvESpirit:
Thanks! Glad you like it.

The code changes are mostly shifting stuff around and updating signatures, nothing major. The change number looks horrid because of all the comment text and whitespacing.

Since I didn't mention this in the original message, the raw assets were to my knowledge lifted from the interior crew recordings of Artemis II and cleaned up. I think the sustain has been created from the same source, but because the crew were using a MIN IMP-like setting at the time (or the attitude adjustments were autonomous), that one is "close enough" - but doctored almost entirely by hand.

Xyon (approved):
Looks good to me. I know you've pulled a build of this off already, and we talked about the sound files themselves becoming part of the licensed work (so essentially given to the project without restriction), so I've no concerns about merging this.

Issue #658, opened 2026-05-18
[XRSound] CTD opening nosecone


danepowell:
Whenever I try to open the nose cone on the Delta Glider (using the keyboard shortcut K or the instrument panel) with XR Sound enabled, I get an immediate crash to desktop. I'll attach the Orbiter log, but there's nothing directly related to the error in it. The event viewer log is below.

I've tried the latest Orbiter x86 and x64 dev builds downloaded from GitHub. I'm playing the Checklist 1 (DG to ISS) scenario. The only module I'm using is XR Sound and the high-res textures for moon and earth. I've played for a couple hours around the Earth and LEO and haven't had any issues prior to this. This only occurs with XR Sound enabled.

[Orbiter.log](https://github.com/user-attachments/files/27941763/Orbiter.log)

```
Faulting application name: Orbiter.exe, version: 0.0.0.0, time stamp: 0x6a055907
Faulting module name: ucrtbase.dll, version: 10.0.26100.8328, time stamp: 0x3e0a6430
Exception code: 0xc0000005
Fault offset: 0x00000000000ed171
Faulting process id: 0x5794
Faulting application start time: 0x1DCE670A17EFB62
Faulting application path: C:\Games\Orbiter-dev\Orbiter.exe
Faulting module path: C:\WINDOWS\System32\ucrtbase.dll
Report Id: 35eb56d5-06fb-4afe-a8f0-21c71aa6dfe5
Faulting package full name: 
Faulting package-relative application ID: 
```

EvESpirit:
I actually fixed this regression previously as it turns out, all the animated things like the retrograde thruster doors and the nosecone and hatch all CTDd the sim as soon as you try. Was some null pointer deref in XRSound.I made a PR. This can be closed when merged.

nadir-line:
@EvESpirit I don't know why but this doesn't compile

PR #660, opened 2026-05-18
[XRSound] Fix CTD (#658)

EvESpirit:
Fixes a null pointer dereference that causes XRSound to hard crash the entire simulation if we try to open the hatch, nosecone or retrograde thrusters.

### DIFF
--- a/Sound/XRSound/src/SoundPreSteps.cpp
+++ b/Sound/XRSound/src/SoundPreSteps.cpp
@@ -224,12 +224,20 @@ void AnimationSoundPreStep::clbkPreStep(const double simt, const double simdt, c
         case AnimationState::StateType::Open:
         case AnimationState::StateType::Closing:
         case AnimationState::StateType::Closed:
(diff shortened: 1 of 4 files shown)


TheGondos:
This is likely a regression with the removal of CString strings. I'll have a look on my side too.
Edit: OK, so GetWavForAnimationState can return nullptr, then the check in LoadAndPlayWavUsingID for !pWavFile.empty() is useless because it's too late.
It can be fixed in AnimationSoundPreStep::clbkPreStep :
```
        case AnimationState::StateType::Opening:
        case AnimationState::StateType::Open:
        case AnimationState::StateType::Closing:
        case AnimationState::StateType::Closed:
		{
			const char *wavefile =  m_animationSounds.GetWavForAnimationState(stateType);
			if(wavefile)
	            LoadAndPlayWavUsingID(m_animationSounds.GetOpenCloseSoundID(), wavefile, false, m_animationSounds.GetPlaybackType());
            break;
		}
        case AnimationState::StateType::Moving:
		{
			const char *wavefile =  m_animationSounds.GetWavForAnimationState(stateType);
			if(wavefile)
	            LoadAndPlayWavUsingID(m_animationSounds.GetMovingSoundID(), wavefile, true, m_animationSounds.GetPlaybackType());
            break;
		}
```

EvESpirit:
I did that now, thanks.

PR #662, opened 2026-05-25
[ExtMFD] Add keyboard input support

EvESpirit:
It's been brought to my attention by rcflyinghokie (@folgers on Discord) that the ExtMFD core module does not support keybinds. I'm not sure if it ever did (as in before ImGUI) but regardless, this PR resolves this issue. ExtMFDs accept keyboard shortcuts if specifically highlighted as this was the cleanest way to do it IMO (as opposed to something like, different modifiers than Shift for each ExtMFD open etc.).

### DIFF
--- a/Src/Orbiter/Mfd.cpp
+++ b/Src/Orbiter/Mfd.cpp
@@ -928,8 +928,8 @@ void Instrument::DisplayModes (int page)
 		skp->SetTextColor (col_green1);
 		name = m->Spec()->name;
 		skp->Text (x, y, name, strlen(name));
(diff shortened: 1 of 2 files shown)


TheGondos:
This works OK. The for loop is slightly offputting but I don't think ImGui gives a choice.
Timed it at less than a microsecond on my hardware anyway so it shouldn't be a problem^^

EvESpirit:
I got rid of the PWR button (superfluous since it's external and able to be X'd out of) and I made sure extmfds can display shortcuts, as per another request by @rcflyinghokie . This is now ready.

PR #663, opened 2026-05-25
[ExtMFD] Fix ExtMFD size amnesia

### LINKED ISSUE #355: Make External MFD save/load a previous set window size
Something that annoyed me for a while now is that the External MFD plug-in doesn't retain its previous set size between uses. Too me on modern 1920x1440 resolution, the standard External MFD window size is just too small to get any information out of it, so I have to resize it every time I bring one up. Also, the shuttle used 1024x1024 MFDs back in 2000 starting with the introduction of the Multi-functional Electronic Display System (MEDS) "glass cockpit" on mission STS-101.

EvESpirit:
This PR fixes #355, an issue with ExtMFD from 3 years ago (sheesh). The core issues were actually two things:

While ImGuiCond_Once *sounds* like the correct flag, it just means something like "apply this rule the first time this code is executed per runtime session.". This obviously causes it to default out. The fix makes it so it only defaults if there's no saved size data for a specific window.

The second one was weird, basically the ImGui dialog's "name" was generated using the memory pointer of the MFDWindow instance - meaning that features like ASLR that are fairly standard these days force it to always think that it's a different window that needs a new name - so eventhough we save (after the first modification), we still are unable to retrieve the size info, so we default. I fixed this with a tiny and simple ID system for extmfd windows since that seemed like the easiest option.

### DIFF
--- a/Src/Plugin/ExtMFD/MFDWindow.cpp
+++ b/Src/Plugin/ExtMFD/MFDWindow.cpp
@@ -20,6 +20,28 @@
 #include "imgui_extras.h"
 #include "IconsFontAwesome6.h"
 
+#include <vector>
+
+static std::vector<bool> g_usedWindowIds;
+
+int GetFreeWindowId() {
+	for (size_t i = 0; i < g_usedWindowIds.size(); ++i) {
+		if (!g_usedWindowIds[i]) {
+			g_usedWindowIds[i] = true;
+			return i + 1;
+		}
+	}
+	g_usedWindowIds.push_back(true);
+	return g_usedWindowIds.size();
+}
+
+void FreeWindowId(int id) {
+	int index = id - 1;
+	if (index >= 0 && index < (int)g_usedWindowIds.size()) {
+		g_usedWindowIds[index] = false;
+	}
+}
+
 // ==============================================================
 // class MFDWindow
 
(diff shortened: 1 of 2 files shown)


Src/Plugin/ExtMFD/MFDWindow.cpp
```diff
@@ -20,6 +20,28 @@
 #include "imgui_extras.h"
 #include "IconsFontAwesome6.h"
 
+#include <vector>
+
+static std::vector<bool> g_usedWindowIds;
+
+int GetFreeWindowId() {
```
TheGondos: could be static

TheGondos:
Gave it a try, this is nice. Also since the position is saved, the layout is conserved too.

Src/Plugin/ExtMFD/MFDWindow.cpp
```diff
@@ -20,6 +20,28 @@
...
+		if (!g_usedWindowIds[i]) {
+			g_usedWindowIds[i] = true;
+			return i + 1;
+		}
+	}
+	g_usedWindowIds.push_back(true);
+	return g_usedWindowIds.size();
+}
+
+void FreeWindowId(int id) {
```
TheGondos: static

EvESpirit:
Did!

PR #664, opened 2026-05-25
[ImGui] Fix debug CMake builds not copying ImGui fonts

### LINKED ISSUE #613: ImGui throws errors due to missing fonts at sim load
When attempting to load any scenario in a Debug build of Orbiter, I get three or four assertion failure popups about ImGui failing to load a font file. I am building from source for x64 Windows, on the latest main branch commit. If I press the "Ignore" button on all of these popups, the sim seems to load okay, but it makes running debug builds of Orbiter frustrating. The errors also don't seem to appear when building for Release, however for all I know the issue is still occurring behind the scenes. I think there may be some missing font files or perhaps a bad path somewhere? @TheGondos

EvESpirit:
Fixes #613. Just modified ImGui's cmakelist to include fonts in debug builds, nothing else.

### DIFF
--- a/Extern/imgui/CMakeLists.txt
+++ b/Extern/imgui/CMakeLists.txt
@@ -11,8 +11,20 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(imgui)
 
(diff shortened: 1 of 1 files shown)


TheGondos:
Are we sure we want to support running outside of the install directory? Sure this change is trivial but there may be other files missing that will cause trouble later.

EvESpirit:
Closing this as per last comment on #613.

PR #665, opened 2026-05-26
[Launchpad] Proper quicksave deletion confirmation

### LINKED ISSUE #600: Remove (or add saveguards) for the "Clear quicksaves" button
This is slightly motivated by spite due to some serious lost work, but I think the "Clear quicksaves" button is both dangerous (it deletes files permanently without any confirmation) and unnecessary (if a user is so determined to erase their quicksaves, they can simply delete the folder). For that reason, I'd like to ask that the button be removed outright, or at least have safeguards put in place, such as a confirmation dialog that defaults to "No" or even having the files be moved to the recycle bin instead of instant deletion.

EvESpirit:
(Hopefully) sufficently fixes #600. The dialogs now look like this:
Orbiter_XTzaRZXzDd
Orbiter_iObMCMtnQ1
image

If something else should be done about this I'd be happy to do it.

### DIFF
--- a/Src/Orbiter/TabScenario.cpp
+++ b/Src/Orbiter/TabScenario.cpp
@@ -666,9 +666,38 @@ INT_PTR CALLBACK orbiter::ScenarioTab::SaveProc (HWND hWnd, UINT uMsg, WPARAM wP
 //-----------------------------------------------------------------------------
 void orbiter::ScenarioTab::ClearQSFolder()
 {
+	fs::path scnpath{ fs::absolute(pLp->App()->ScnPath("Quicksave")) };
+	scnpath.replace_extension(); // remove ".scn"
+
+	std::string msg = "Are you sure you want to delete all quicksaves? This affects:\n";
+	int qsCount = 0;
+	
+	if (fs::exists(scnpath) && fs::is_directory(scnpath)) {
+		for (auto& entry : fs::directory_iterator(scnpath)) {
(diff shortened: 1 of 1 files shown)


Src/Orbiter/TabScenario.cpp
```diff
@@ -666,9 +666,38 @@ INT_PTR CALLBACK orbiter::ScenarioTab::SaveProc (HWND hWnd, UINT uMsg, WPARAM wP
...
+	}
+	
+	if (MessageBox(LaunchpadWnd(), msg.c_str(), "Clear Quicksaves", MB_YESNO | MB_ICONWARNING) != IDYES) {
+		return;
+	}
+
 #ifdef _WIN32
 	// SHFileOperation needs an absolute path
-	fs::path scnpath{ fs::absolute(pLp->App()->ScnPath("Quicksave")) };
 	scnpath.replace_extension(); // remove ".scn"
```
TheGondos: is it still necessary?

Src/Orbiter/TabScenario.cpp
```diff
@@ -682,7 +711,6 @@ void orbiter::ScenarioTab::ClearQSFolder()
 		fs::create_directory(scnpath);
 	}
 #else
-	fs::path scnpath{ pLp->App()->ScnPath("Quicksave") };
 	scnpath.replace_extension(); // remove ".scn"
```
TheGondos: idem

EvESpirit:
I've removed the entire block.

PR #667, opened 2026-06-19
[Core] Fix 20-year old ephemeris interpolation bug

EvESpirit:
This small commit implements a Cubic Hermite Spline ephemeris method. The old method failed to correct for the velocity vector after adjusting the planet's path. It basically drew a straight line from point A to point B on a ellipse (which itself was done to save performance), cutting the corner, and after pushing the position of the body outwards to create the curved path it needed, it just left the velocity pointing along the original straight path. The CHS fixes this by taking into account all four pieces of information and "bending a smooth curve" between them. This mathematically guarantees that any attempts to calculate speed along any point of our curve will perfectly match the interpolated velocity variable.

Why is this important?

Because, while "high gravity" bodies naturally hide the consequences of this bug by essentially shrinking the size of the velocity error by having "close enough to a straight line" paths through space between interpolation points. Even on certain bodies where we would expect this behavior to kick in, the friction capacity of most vessels still hid this bug due to the bodies' inherently greater gravity, giving the ship massive grip that allowed fullstop landings anyway. So this bug really only showed up on *very* low gravity bodies with tight orbits, like Mimas.

Without this fix, full-stop landings on these bodies are simply impossible, because you will always be sliding along the surface through no fault of your own. For Mimas, this value comes out at exactly 0.17m/s. That's not insignificant and looks really jarring. Especially once we reach another interpolation point, the sliding vessel appears to "bounce" off of nothing in a completely random direction at the same velocity.

### DIFF
--- a/Src/Celbody/Galsat/Galsat.cpp
+++ b/Src/Celbody/Galsat/Galsat.cpp
@@ -107,6 +107,17 @@ inline double Radius (double *data)
 	return sqrt (data[0]*data[0] + data[1]*data[1] + data[2]*data[2]);
 }
 
+// ===========================================================
+// Name: Interpolate()
+// Desc: Interpolates position and velocity using a Cubic 
+//       Hermite Spline. Replaces the old linear interpolation 
+//       method that was causing the interpolated velocity to 
+//       deviate from the derivative.
+//       Values returned in 'data' are the interpolated
+//       position (data[0..2]) and velocity (data[3..5]).
+//       't' is the time to interpolate at, bounded by
+//       samples 's0' and 's1'.
+// ===========================================================
 void Interpolate (double t, double *data, const Sample *s0, const Sample *s1)
 {
 	int i;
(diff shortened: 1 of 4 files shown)


Xyon:
I'm not particularly familiar with the maths involved here so I can't really comment on the implementation's suitability (I'm sure you've tested it and that'll suffice for me), but I do think the code would benefit from a little clarity about what it's doing - as written it just dives right into a bunch of opaque maths and some number constants seem to arrive out of nowhere. Could you perhaps add a comment to note why that's happening, what end values the calculations are trying to arrive at, and specifically that it's a CHS implementation? for future lost scientists scratching their heads (including me)

EvESpirit:
Entirely fair, fixed it now.

TheGondos:
This looks good to me, CHS are good when you need C1 continuity as is the case here.
A question : did you try removing fast ephemeris entirely? I have no idea how current hardware would handle it.

EvESpirit:
I wanted to mess with the existing code as little as possible, because the whole ephemeris system in basically it's entirety is sort of going away, pending Matthew's work. In the same vein, I also had some qualms about code duplication that was really caused by some ancient decision by Martin, but all of that is just a bunch of work I didn't want to do.

I do reckon that current hardware would handle it though. I just didn't want to do more work that would potentially be negged by the SPICE PR. When time comes to merge it I'll come back to review how we handle things. I don't know enough about SPICE to make that decision now.

n7275:
> This looks good to me, CHS are good when you need C1 continuity as is the case here. A question : did you try removing fast ephemeris entirely? I have no idea how current hardware would handle it.

You can, in most cases change the interpolation at a user-level by reducing the interpolation interval in the celbody config. You will usually gain less than a meter of accuracy. fastEphemeris is really useful for precomputing a bunch of (time, position, vel) arrays by external modules (trajectory planning tools, etc). Would definitely leave that as is.

n7275:
> I wanted to mess with the existing code as little as possible, because the whole ephemeris system in basically it's entirety is sort of going away, pending Matthew's work. In the same vein, I also had some qualms about code duplication that was really caused by some ancient decision by Martin, but all of that is just a bunch of work I didn't want to do.
> 
> I do reckon that current hardware would handle it though. I just didn't want to do more work that would potentially be negged by the SPICE PR. When time comes to merge it I'll come back to review how we handle things. I don't know enough about SPICE to make that decision now.

In the Spice PR (and all the hard work was done by Ajaja, not me). clbkEphemeris is just calling this function: https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/spkgeo_c.html and the internal spice functions do the Chebyshev polynomial magic. It's not alltogether different than the way the VSPO87 modules are implemented, except for the fact that there are not seperate overloaded classes/modules for each celbody, they are all just an instance the same class in spice.dll [or .so some day]

I also added a CMAKE option so that Orbiter can be build with the legacy VSOP87/ELP82/Galsat/Satsat etc modules, or with the spice modules.

EvESpirit:
Oh, that's perfect then. Thanks for the explanation. At least this code will stay useful despite SPICE being the preffered solution going forward.

PR #669, opened 2026-06-27
Add support to render surface labels to custom cameras

M3d1cZ4pp3r:
This changes the Custom Camera render path to support rendering surface labels when a respective flag is set at camera creation.
Additionally, a scale can be set from the gcCore API. This is needed because we don't know how the target surface will be scaled on the actual screen later on, so the client of the custom camera needs to scale up the surface labels so they are readable.
The scale is applied to the symbols as well as on the fonts. For the fonts a cache of bigger fonts is created and used.
While there are other solutions that work and might take less resources, e.g. I did a small change to render the text to a sprite before rendering it to the sketchpad. This makes it sensitive to the sketchpads transform matrix and the font is upscaled too.
 
But this is a more invasive approach into the general text render path, while my requested approach only changes Custom Camera and Label Rendering logic, on the cost of keeping bigger fonts in memory and rendering them.

Additionally the Camera structures need to carry that scale information and also the dimensions of the render target.
This is because the projection logic of the Label Renderer needs to know the viewport dimensions.

In general I tried to minimize invasiveness and kept it backwards compatible, so no interface should break.

The reason for this change is that in the NASSP Apollo Project the superimposition technique on the optics sextant uses a custom camera. The user cannot spot landmarks then anymore. This allows to still render them, and resize them according to the size of the custom camera rendered texture on the screen.

### DIFF
--- a/OVP/D3D9Client/Scene.cpp
+++ b/OVP/D3D9Client/Scene.cpp
@@ -3145,10 +3145,6 @@ void Scene::ExitGDIResources ()
 	oapiReleaseFont(pAxisFont);
 	oapiReleaseFont(pLabelFont);
 	oapiReleaseFont(pDebugFont);
-
-	for (int i = 0; i < 4; ++i) {
-		gc->clbkReleaseFont(label_font[i]);
-	}
 	
 	// Only delete the cache. The label_font are just weak refs!
 	for(auto &[size, font] : labelFontCache) {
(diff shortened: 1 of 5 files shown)


spacex15:
#include <algorithm> is missing in scene.cpp to use std::clamp

causes a compilation error in visual studio

Xyon:
> #include <algorithm> is missing in scene.cpp to use std::clamp
> 
> causes a compilation error in visual studio 

fixed in #675

